PR #6, opened 2019-09-18
tomislav changes

forgedisaman:
updated version of Tomislav with faster switch speed with a debuff on health given from medics
and removing the old 40% accuracy as it's already in the base game.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -497,7 +497,7 @@
 		}
 		"424"	//Tomislav
 		{
-			"desp"			"Tomislav: {positive}+25% faster switch speed, 25% extra spin up time, {negative}-70% health from healers"
+			"desp"			"Tomislav: {positive}+25% extra switch speed, 25% extra spin up time, {negative}-70% health from healers"
 			"attrib"		"87 ; 0.5 ; 178 ; 0.5 ; 69 ; 0.3"
 		}
 		"811"	//Huo-Long Heater
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -497,8 +497,8 @@
 		}
 		"424"	//Tomislav
 		{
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo:
One small problem i see:
All Miniguns stat:
```"desp"		"Primary: {positive}25% faster weapon switch and 25% faster spin up time" ```
New Tomislav stat:
```"desp"			"Tomislav: {positive}+50% faster switch speed, 25% extra spin up time, {negative}-70% health from healers"```

Both have `faster switch speed` and `extra spin up time` listed, but `+50% faster switch speed` states all miniguns don't have one, while `25% extra spin up time` states there already one.
Should be worded `+25% extra switch speed` instead.

**Comment by FortyTwoFortyTwo** [comment]
Now i think more about it, health from healers penalty actually don't need that much of a huge nerf, and accuracy removal could've been spin up time removal instead.
Lets see how this new Tomislav goes first.

File: addons/sourcemod/configs/vsh/vsh.cfg [hunk 1]
```diff
@@ -497,8 +497,8 @@
 		}
 		"424"	//Tomislav
 		{
-			"desp"			"Tomislav: {positive}+40% more accurate, 25% extra spin up time"
-			"attrib"		"106 ; 0.6 ; 87 ; 0.5"
+			"desp"			"Tomislav: {positive}+25% extra switch speed, 25% extra spin up time, {negative}-70% health from healers"
```
shinsasch: shouldn't '25% extra spin up time' be '+25% faster spin up time'?

woisalreadytaken: in this case, it's applying a 25% faster spin up time on top of an already existing 25% faster spin up time (given as an attribute to all heavy primaries, also explains why it's at 0.5). 

while "extra" was likely added in good spirit to help clear up that it's a bonus, you're right that it would mean slower, but just having `+25% faster spin up time` may confuse players who are reading the description since all miniguns already have it.

maybe combining your suggestion with the current description with something like "Additional/Extra 25% faster spin up time" would clear it up better

PR #9, opened 2019-09-19
Update vsh.cfg

forgedisaman:
updated text on Tomislav from extra to additional to make the upside clearer.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -497,8 +497,8 @@
 		}
 		"424"	//Tomislav
 		{
-			"desp"			"Tomislav: {positive}+25% additional switch speed, 40% additional accuracy {negative}-60% overheal"
-			"attrib"		"87 ; 0.5 ; 106 ; 0.6 ;  105 ; 1.6"
+			"desp"			"Tomislav: {positive}Additional +25% faster switch speed, additional 20% accuracy, {negative}-60% max overheal"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -497,7 +497,7 @@
 		}
 		"424"	//Tomislav
 		{
-			"desp"			"Tomislav: {positive}+25% extra switch speed, 25% extra spin up time, {negative}-70% health from healers"
+			"desp"			"Tomislav: {positive}+25% additional switch speed, 25% extra spin up time, {negative}-70% health from healers"
```
FortyTwoFortyTwo: Would also change `25% extra spin up time` to `additional 25% faster spin up time`.
Also `additional 25% faster weapon switch` would be better to match same with spin up?

shinsasch:
> I still want to hear what others think about current tomislav stats, whenever if it okay or should be changed

👍 


An idea I have would be to replace the lower healing received attribute with the `patient overheal penalty` attribute to give the wearer max 366 (47%, very close to two hits of a normal boss) or 375 HP (50%) overheal, but the lower healing received is fine nontheless.

forgedisaman:
> An idea I have would be to replace the lower healing received attribute with the patient overheal penalty attribute to give the wearer max 366 (47%, very close to two hits of a normal boss) or 375 HP (50%) overheal, but the lower healing received is fine nontheless.

yeah this seems like a better idea than the healing rate penalty without affecting other heavy gameplay +1 to that.

PR #10, opened 2019-09-19
family buisness changes

forgedisaman:
speed on hit, lower reload speed. hit and run style weapon idea.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -519,6 +519,11 @@
 		{
 			"prefab"		"159"
 		}
+		"425"	//The Family Business
+		{
+			"desp"			"The Family Business: {positive}speed boost on hit, {negative}50% slower reload speed"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -519,6 +519,11 @@
 		{
 			"prefab"		"159"
 		}
+		"425"	//The Family Business
+		{
+			"desp"			"The Family Business: {positive}speed boost on hit, {negative}50% slower reload speed"
```
FortyTwoFortyTwo: `The` in weapon name can drop out like every other weapons, also `Speed boost` with S caps lock

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -519,6 +519,11 @@
...
+			"attrib"		"737 ; 1 ; 96 ; 1.5"
```
FortyTwoFortyTwo: Small nitpicking, value `1` should be in float `1.0` to keep it consistent with other weapons/attrib

PR #19, opened 2019-10-04
New HHH Ghost Rage

FortyTwoFortyTwo:
This is something i meant to finish a while ago, but here it is, replaces default scare rage.

When rage is used, HHH turns into a ghost for 5 seconds (10 for super rage), can't take any damage and can fly, but can't attack.
When nearby players, medigun beam particle effect appears from victim to ghost, and steals health. Steals 20hp per sec while gains 40hp per sec, which is doubled of stolen health. Because of health gain, HHH's base health is reduced to same as Gentle Spy, and rage requirement is increased from 2500dmg to 3000dmg.

Along with this, nearby players also get "spooky" effects at random, this includes:
- Purple fade with thunder noises
- Teleport-swap to another nearby player
- Force switched to random weapon
- Eye angle view force changed to random direction

### DIFF
--- a/addons/sourcemod/scripting/saxtonhale.sp
+++ b/addons/sourcemod/scripting/saxtonhale.sp
@@ -3041,7 +3041,7 @@ stock int TF2_SpawnParticle(char[] sParticle, float vecOrigin[3] = NULL_VECTOR, 
 		AcceptEntityInput(iParticle, "Start");
 	}
 	
-	//Really only need ref of entity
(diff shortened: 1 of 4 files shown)


FortyTwoFortyTwo:
Something to do before merge: Update HHH's info for new rage, and check if spooky effects work properly, as it only tested on bots so far.

FortyTwoFortyTwo:
This should be good to go now

PR #22, opened 2019-10-05
Make Demopan's charge instakill

FortyTwoFortyTwo:
Because Demopan's rage will still suck after PR #21 

This is quite of a hacky workaround inorder to make this work, because demoshield only deal damage if charge ends, while rage force charge even after bash.

Whenever demopan touches enemy player, call `SDKHooks_TakeDamage` a frame later to prevent possible crash, and update `player_death` event to make kill icon etc look like a charge kill.


FortyTwoFortyTwo:
One thing im consider to do aswell, is having charge bash instakill sentry/buildings, esp when sentries can aimbot damage demopan while running at sanic speed with no resists

FortyTwoFortyTwo:
When charge instakilling buildings is tested, kill icon shows as a generic skull while `object_destroyed` event doesn't even get called to modify, and even then i'm not sure if there anyway to fix with event.
Will have to leave that as wont fix as it just a kill icon anyway.

PR #23, opened 2019-10-06
Fix Brave Jump ability not playing VO

woisalreadytaken:
The functions update made voicelines from brave jumps not play anymore, this pull request simply fixes that.


woisalreadytaken:
Done, didn't bother checking if abilities that don't play any sounds in-game actually tried to play sounds if possible

PR #27, opened 2019-10-08
Introduce Function Tags

FortyTwoFortyTwo:
There a lot of changes with tags, this may be a bit confusing to some ~~or if anyone even cares about this~~.

//---//

Lets take a look at new buff banner
```
"129"	//Buff Banner
{
	"desp"			"Buff Banner: {positive}Barrage of minicrit rockets on use"
	
	"banner"			/List of functions called when client use banner
	{
		"Tags_AddAttrib"
		{
			"target"		"primary"	//Primary weapon
			"index"		"6"		//Faster firing rate
			"value"		"0.1"	//10x faster firing speed
			"duration"	"5.0"	//After 5 sec, remove attrib
		}
		
		"Tags_AddAttrib"
		{
			"target"		"primary"	//Primary weapon
			"index"		"869"	//Crit becomes minicrit
			"value"		"1.0"	//true
			"duration"	"5.0"	//After 5 sec, remove attrib
		}
		
		"Tags_AddClip"
		{
			"target"		"primary"	//Primary weapon
			"amount"		"16"		//Adds 16 clip
			"duration"	"5.0"	//After 5 sec, remove any extra clips
		}
	}
}
```
First thing we have is `desp` and `banner`, new `banner` acts like a tag with list of functions to call. Whenever client uses buff banner, all functions listed will be called with param(s). Multiple params allow us to make this much more customizeable, and more easier to understand what each values does, i guess?

//---//

Another cool thing we have is `filter`, lets look at enforcer with crits while disguised
```
"460"	//Enforcer
{
	"desp"			"Enforcer: {positive}Receive crits while disguised, {negative}critical damage is affected by range, no minicrits"
	"attrib"		"868 ; 1.0"
	"minicrit"		"0"
	
	"think"
	{
		"filter"
		{
			"cond"	"3"	//Requires disguise
		}
		
		"Tags_AddCond"
		{
			"cond"	"56"	//Give crits
		}
	}
}
```
There a function `Tags_AddCond` with crits to be called, however there `filter` with `cond` as 3 (`TFCond_Disguised`). This means `Tags_AddCond` will only be called if player is disguised, allowing lots of many different combinations with filter.

//---//

I could talk further on other stuffs, but i'll let you figure out yourself in new `vsh.cfg`, or if you want to say any thoughts on this before i go further into adding any missing bits, cleanups and fixes.


FortyTwoFortyTwo:
This is probably one of the toughest thing i've done, the amount of enum struct, methodmap and handles been used, but it nearly done. Just missing some calls and possible bugs, then should be good to go after this. This probably took too much effort in the end, but oh well.

Quick summary on how each files/methodmap work in code for nerds:

`tags_core.sp` - The "main" area, `Tags` with list of every tags and functions, along with starting a call on which functions, depending on class/slot/filter/block.
`tags_filters.sp` - `TagsFilter` methodmap used to store filter stuffs, and whenever if allowed to pass or not.
`tags_function.sp` - `TagsFunction` methodmap to store function to call, and points to which params/block/override used for this function.
`tags_params.sp` - `TagsParams` methodmap to store every params used from function and passing around to use.
`tags_block.sp` - `TagsBlock` methodmap to store every `Tags` id attempting to block this function, whenever if allowed to be blocked or not.
`tags_name.sp` - Sync function names from block and overrides after everything loaded, to make plugins run faster when calling.
`tags_call.sp` - List of available calls, and actually calling a function.
`tags_target.sp` - List of available targets, and getting player/entity target depending on type given.
`tags_damage.sp` - Manages player damages with params entered/given, and changing values.
`tags.sp` - Freely every function callbacks.

FortyTwoFortyTwo:
Pray i didn't missed anything and not breaking everything horribly, lets do this.

PR #33, opened 2019-10-21
Update spcomp public methodmap compiler

FortyTwoFortyTwo:
3 things changed to spcomp:
- Update SourceMod version from 1.10.6442 to 1.10.6453, only SourcePawn related change is adding compile warning to legacy enum struct. 1.10.6453 is also the git version declared to be stable.
- Update compile caption to display `SourcePawn Public Methodmap Compiler 1.10.6453`, to easily tell whenever if compiling with normal compiler, or custom compiler
- Added new constant `__sourcepawn_methodmap__` to spcomp, this allows `saxtonhale.sp` to throw a compile warning if not compiling with custom compiler.

Also renamed "custom" to "SourcePawn Public Methodmap", as that sounds better and gives more info on differences.


compiler/compile-1.10.cpp
```diff
@@ -1139,6 +1139,7 @@ setconstants(void) {
 
     add_constant("__Pawn", VERSION_INT, sGLOBAL, 0);
     add_constant("__LINE__", 0, sGLOBAL, 0);
+	add_constant("__sourcepawn_methodmap__", 1, sGLOBAL, 0); //Public methodmap: Add new constant
```
FortyTwoFortyTwo: 🍉

PR #41, opened 2019-10-25
Fix Glitched Robot not playing regular and jump VO

woisalreadytaken:
Pretty simple. The code for replacing default voicelines was missing a replacement step and jump voicelines weren't precached.

Since we're here, might as well update the rage description.

### DIFF
--- a/addons/sourcemod/scripting/vsh/bosses/boss_demorobot.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_demorobot.sp
@@ -161,14 +161,12 @@ methodmap CDemoRobot < SaxtonHaleBase
 			
 			char file[PLATFORM_MAX_PATH];
 			strcopy(file, PLATFORM_MAX_PATH, sample);
-			ReplaceString(file, sizeof(file), "vo/", "vo/mvm/norm/", false);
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/vsh/bosses/boss_demorobot.sp
```diff
@@ -156,18 +156,25 @@ methodmap CDemoRobot < SaxtonHaleBase
 	{
 		if (strncmp(sample, "vo/", 3) == 0)
 		{
+			if (StrContains(sample, "vo/mvm/", false) == 0)
+				return Plugin_Continue;
+			
 			char file[PLATFORM_MAX_PATH];
 			strcopy(file, PLATFORM_MAX_PATH, sample);
 			ReplaceString(file, sizeof(file), "vo/", "vo/mvm/norm/", false);
+			ReplaceString(file, sizeof(file), "demoman_", "demoman_mvm_", false);
```
FortyTwoFortyTwo: Would be quicker if single `ReplaceString` from `vo/demoman_` to `vo/mvm/norm/demoman_mvm_`, unless if that doesn't work on voicelines i do not know of

PR #42, opened 2019-10-26
Airblast: Replace ammo cost to damage requirement

FortyTwoFortyTwo:
VSH with airblast issue will never end.

Currently, 70 ammo cost for airblast still isn't enough to help prevent spam, because of dispenser able to keep feeding ammo endless amount, along with ammo packs around the map. Thermal Thruster also stacks up annoyance with escaping and landing to dispenser/ammo pack to keep up the annoyance.

Instead of using ammo as a cost, 500 damage is required everytime airblast is used, with no ammo cost. Hud will display percentage progress on when able to use airblast from damage dealt. Boss's airblast reduction in airblast vulnerability has been removed along with that.

Hopefully this should make airblast feels more rewarding.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -200,7 +200,7 @@
 			"Primary"
 			{
 				"desp"		"Primary: {positive}+200% afterburn damage bonus, {neutral}replace airblast ammo cost to 500 damage requirement"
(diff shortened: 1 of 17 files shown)


FortyTwoFortyTwo:
Discovered myself a bug, 100% airblast is lost when switching weapon then back to flamethrower.
This PR shouldn't be merged until said is fixed

PR #44, opened 2019-10-27
Fix high environmental damage teleporting bosses once per valid spawn point and add the stun animation to it

woisalreadytaken:
High environmental damage has been stunning bosses for a while but it just doesn't let bosses move instead of showing them they're actually stunned, so I changed that to give better visual feedback to them.
On the way, I found out that bosses get teleported and stunned multiple times if there's more than one spawn point for their team, so I added a check to stop the plugin from doing so if they were already teleported (and to also stop it from fucking destroying people's eardrums with the new stun sound).

### DIFF
--- a/addons/sourcemod/scripting/vsh/base_boss.sp
+++ b/addons/sourcemod/scripting/vsh/base_boss.sp
@@ -327,7 +327,7 @@ methodmap SaxtonHaleBoss < SaxtonHaleBase
 							float vecNoVel[3];
 							TeleportEntity(this.iClient, vecPos, NULL_VECTOR, vecNoVel);
 							damage = (damagetype & DMG_ACID) ? this.flEnvDamageCap/3.0 : this.flEnvDamageCap;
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/vsh/base_boss.sp
```diff
@@ -325,7 +327,8 @@ methodmap SaxtonHaleBoss < SaxtonHaleBase
 							float vecNoVel[3];
 							TeleportEntity(this.iClient, vecPos, NULL_VECTOR, vecNoVel);
 							damage = (damagetype & DMG_ACID) ? this.flEnvDamageCap/3.0 : this.flEnvDamageCap;
-							TF2_StunPlayer(this.iClient, 2.0, 1.0, 257, 0);
+							TF2_StunPlayer(this.iClient, 2.0, _, TF_STUNFLAGS_NORMALBONK, 0);
+							bTeleported = true;
```
FortyTwoFortyTwo: `break` more simpler to escape while loop, without needing `bTeleported` variable

PR #45, opened 2019-10-27
Balance: Demoshield rework

woisalreadytaken:
Demoknights are pretty eh, some people say they're annoying but the only viable set is tide turner + eyelander. The breaking mechanic makes it so it's easier for a demoknight to snowball ahead if they're using the eyelander (and with that become very annoying to fight against later on) but it just makes the class boring if they're using any other melee. 

The only viable shield is the tide turner because in this gamemode, specific damage types are uncommon to come and being able to turn while charging is far better than anything else.

This PR, thanks to 42's new tag stuff, gives a bit of life to shields that aren't the tide turner and replaces the breaking mechanic, changes:
- The breaking mechanic is REMOVED (except for the razorback). Instead, players using shields will take full damage on the first hit and have their charge meter fully depleted on any melee hit instead. Player using boots without the eyelander will now die in 2 hits and in 1 otherwise.

- The chargin' targe will give a defense buff for 6 seconds when bashing the boss with it, it doesn't scale with range but it does have a minimum range threshold. Useful with a medic or half-zatoichi!

- The splendid screen has increased damage (from +70% to +200%) and knockback by a massive amount, damage and knockback scale with range. Useful with the persian persuader!

- The tide turner has no changes specific to it, as turning when charging is a very good upside on its own. Useful for hybridknights!


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1041,12 +1043,48 @@
...
+				"params"
+				{
+					"damagetype"	"knockback"
+				}
+				
+				"Tags_AddAttrib"
+				{
+					"target"		"victim"
+					"cond"			"127"
+					"duration"		"0.5"
```
FortyTwoFortyTwo: wasnt airblast cond suppose to be perma until touches ground, or was that too much

woisalreadytaken: nah i just forgot to update that on my config loool

FortyTwoFortyTwo: Should add comment to note duration is perma until touches ground

PR #43, opened 2019-10-27
Last Yeti

Mikusch:
Related to #40 

The easiest stuff is done. Need to add ability and rage and adjust sounds so they aren't overly obnoxious.

PR #53, opened 2019-11-03
Add bonemerge model boss ability

FortyTwoFortyTwo:
This took quite a bit of hacky and pain to get this working. Instead of using `player` entity as a model, we use `prop_dynamic_override` entity parented to player, along with `EF_BONEMERGE` where entity's model would follow player's model animation. Two `prop_dynamic_override` is needed, entity 1 parented to entity 2 parented to player. I tried to use just 1 of em, but it just doesn't work.

This is still work in progress. HHH's model need to be resided to more reasonable size, and probably broke to shit with ghost rage.

`CModelOverride` ability can also be used for Yeti (#43) and Merasmus (#51), using TF2's model without needing custom model.

### DIFF
--- a/addons/sourcemod/scripting/saxtonhale.sp
+++ b/addons/sourcemod/scripting/saxtonhale.sp
@@ -158,6 +158,25 @@ enum
 	LAST_SHARED_COLLISION_GROUP
 };
 
+// entity effects
+enum
+{
+	EF_BONEMERGE			= (1<<0),	// Performs bone merge on client side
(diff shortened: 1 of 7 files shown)


FortyTwoFortyTwo:
Still can't find myself a way to make HHH model look nice to fit with player's scale. Using bonemerge cause HHH model to look really crunched, large model crammed to player's bone. Changing prop's model scale doesn't help anywhere as bonemerge prevents it.

FortyTwoFortyTwo:
Decided to merge this somewhat soon, but without HHH's default model due to size. Really just so PR #43 can use this, yeti size pretty much same as heavy.

PR #54, opened 2019-11-03
Fix Seeman's super rage sometimes nuking himself

woisalreadytaken:
The code returns `Plugin_Stop` to avoid this, but it doesn't work all (or maybe even most of) the time. Changing self-damage to 0.0 and returning `Plugin_Changed` instead seems to consistently work.

### DIFF
--- a/addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp
@@ -53,7 +53,13 @@ methodmap CSeeMan < SaxtonHaleBase
 		char sWeaponClassName[32];
 		if (weapon >= 0) GetEdictClassname(inflictor, sWeaponClassName, sizeof(sWeaponClassName));
 		
-		if (this.iClient == attacker && strcmp(sWeaponClassName, "tf_generic_bomb") == 0) return Plugin_Stop; // Don't let the bombs from the bomb ability damages us!
+		//Disable self-damage and knockback from bomb rage ability
+		if (this.iClient == attacker && strcmp(sWeaponClassName, "tf_generic_bomb") == 0)
+		{
+			damage = 0.0;
+			damagetype = DMG_PREVENT_PHYSICS_FORCE;
+			return Plugin_Changed;
+		}		
 
 		EmitSoundToAll(SEEMAN_SEE_SND, this.iClient, SNDCHAN_VOICE, SNDLEVEL_SCREAMING);
 		return Plugin_Continue;
--- a/addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp
@@ -51,15 +51,10 @@ methodmap CSeeMan < SaxtonHaleBase
 	public Action OnTakeDamage(int &attacker, int &inflictor, float &damage, int &damagetype, int &weapon, float damageForce[3], float damagePosition[3], int damagecustom)
 	{
 		char sWeaponClassName[32];
-		if (weapon >= 0) GetEdictClassname(inflictor, sWeaponClassName, sizeof(sWeaponClassName));
+		if (inflictor >= 0) GetEdictClassname(inflictor, sWeaponClassName, sizeof(sWeaponClassName));
 		
-		//Disable self-damage and knockback from bomb rage ability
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp
```diff
@@ -53,7 +53,13 @@ methodmap CSeeMan < SaxtonHaleBase
 		char sWeaponClassName[32];
 		if (weapon >= 0) GetEdictClassname(inflictor, sWeaponClassName, sizeof(sWeaponClassName));
 		
-		if (this.iClient == attacker && strcmp(sWeaponClassName, "tf_generic_bomb") == 0) return Plugin_Stop; // Don't let the bombs from the bomb ability damages us!
+		//Disable self-damage and knockback from bomb rage ability
+		if (this.iClient == attacker && strcmp(sWeaponClassName, "tf_generic_bomb") == 0)
+		{
+			damage = 0.0;
+			damagetype = DMG_PREVENT_PHYSICS_FORCE;
```
FortyTwoFortyTwo: This shouldn't be needed. TF2 rounds float damage to int, and if damage is 0 the hurt function never get called with whatever damagetype.

Kenzzer: `Plugin_Handled` & `Plugin_Stop` both prevents the function OnTakeDamage to run through, as demonstrated here:
https://github.com/alliedmodders/sourcemod/blob/cb8d92edebafafc7056e593fd5c96947dbcdab62/extensions/sdkhooks/extension.cpp#L1156

The above code wouldn't change the current behaviour.
https://sm.alliedmods.net/new-api/core/Action

If damage immunity doesn't work as claimed in OP, then the if statement is probably failing, like inflictor class perhaps not filled and/or changed by another plugin before it reached there.

FortyTwoFortyTwo: Others rarely had problems returning `Plugin_Stop` not helping it, it's very weird.
Along with testing `Plugin_Stop` myself, found out bomb/nuke `weapon` entity is invalid (`-1`) so no classname. So that's actually where the root of the problem.

Kenzzer: `Plugin_Stop` very much work as intended, if it doesn't then it's some plugin's fault no doubt there.

https://github.com/redsunservers/VSH-Rewrite/blob/7a7e8a441f1813ea4d184b973f85cee5fe3a8cec/addons/sourcemod/scripting/vsh/bosses/boss_seeman.sp#L54

But the problem is very clear anyways. The if statement is failing, just has to be replaced with `if (inflictor >= 0)` instead of `if (weapon >= 0)` since as you said weapon field is not filled for tf_generic_bomb dmg.

woisalreadytaken: Yeah, turns out what I had here wouldn't change anything, I just happened to never get killed when spamming rages after changing it because the über duration of the rage sometimes doesn't match the last explosion's timing, simply put

FortyTwoFortyTwo: You tested to be very sure inflictor is the bomb?

woisalreadytaken: yes
image

Kenzzer: If that comfort epicfighter42 ;) more, yeah also using it for a boss of mine works fine.
Besides I made that oversight when the mod was in alpha I should be apologizing for it, so yeah whoops sorry I didn't test correctly with Alex.

FortyTwoFortyTwo: I'm fine with this, just want to make sure it really works. Though i'll still be confused when some claims `Plugin_Stop` doesn't work when i can't really see a reason why, likely just bad code in the end 🍉

Kenzzer: It could only break if the extension does a bad job at it, but SM is consistent on their use of action value. It's worth doubting when you deal with a non SM ext.

Though Sourcemod is quite deceptive on a few things, like `SDKHook_OnTakeDamage_Alive`  it's written as not taking an `Action` return value, only `void` and yet accept it due to using the same code as `SDKHook_OnTakeDamage`.

PR #55, opened 2019-11-06
Eviction Notice

StarInAPond:
Currently EN is too good for escaping rather then for taking a high ground, — reason why 50% were changed to 75% in the past, I suggest reverting this change, we already have GRU for dodging punches.


FortyTwoFortyTwo:
Can't tell much about this as i haven't been annoyed too much, but looks okay to me when also removing perma health drain

PR #56, opened 2019-11-06
Re-add no knockback while boss on ground

FortyTwoFortyTwo:
Was accidentally removed in massive PR #27, oops!


FortyTwoFortyTwo:
Just realized that the no-knockback only affects if either on ground, or crouching. Meaning you can in theory never take generic knockbacks by crouching while jumping/superjump. This is even in VSH-Legacy, doesn't feel right to me.

One option is to remove the crouch thing, and bringing back super weightdown. Though this may still be an issue if trying to kill someone with knockback at very small high spot? Another option is to directly remove generic knockback, but that's not fun.

FortyTwoFortyTwo:
Going for weightdown route, removed crouch no-knockback. Weightdown will be added in separate PR

PR #61, opened 2019-11-10
Disable Ice modifier

woisalreadytaken:
There have been complaints about Ice modifier not being a good modifier as the upside wasn't noticeable, so people were deeming it a straight downgrade.
...well, that's because it apparently is.

The modifier's upside wouldn't work because it was modifying the victim's speed, as intended, but would try to recalculate the boss' speed instead. This has been "fixed".

That's not enough, as the function that calculates max speed uses a pre hook, rather than post. It'll always return `0.0` and make a player move at 520 hu/s if their speed multiplier is modified. 42 and I looked into this for a bit, but no matter what was done it'd crash the server, so I suggest that the modifier should be disabled for now. The modifier will still be available for server administrators to force it upon players, making it sort of a challenge modifier.

### DIFF
--- a/addons/sourcemod/scripting/vsh/modifiers/modifiers_ice.sp
+++ b/addons/sourcemod/scripting/vsh/modifiers/modifiers_ice.sp
@@ -9,6 +9,11 @@ methodmap CModifiersIce < SaxtonHaleBase
 	{
 	}
 	
+	public bool IsModifiersHidden()
+	{
+		return true;
+	}
+	
 	public void GetModifiersName(char[] sName, int length)
 	{
 		strcopy(sName, length, "Ice");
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo:
Gameplay issue is also similar to Jump modifier, needing superjump to actually have a change, expect it just harder to do fall damage as non-jump boss.

If there no other way to fix this and turning into a challenge modifier, I would've moved code to private repo to be redsun-exclusive. This means delete ice modifier in this repo and speed detour that never worked anyway.

woisalreadytaken:
Closing this because of PR #62, Something must've been done to the slowdown stunflag so diagonal movement isn't as much of a pain since 42 changed the ice effect to its current state, and that's amazing

PR #62, opened 2019-11-10
Fix ice modifier

Mikusch:
We now simply use the native `TF2_StunPlayer` provided in `tf2.inc` to both apply and remove the slowdown effect.

### DIFF
--- a/addons/sourcemod/scripting/vsh/modifiers/modifiers_ice.sp
+++ b/addons/sourcemod/scripting/vsh/modifiers/modifiers_ice.sp
@@ -1,5 +1,6 @@
 #define ICE_SLOWDOWN 0.7
 #define ICE_DURATION 2.0
+#define ICE_RANGE 250.0
 
 static float g_flClientIceSlowdown[TF_MAXPLAYERS+1];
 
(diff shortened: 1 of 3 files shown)


addons/sourcemod/scripting/vsh/modifiers/modifiers_ice.sp
```diff
@@ -124,8 +119,5 @@ public Action Timer_ResetSpeed(Handle hTimer, int iRef)
 	if (iClient <= 0 || iClient > MaxClients || !IsClientInGame(iClient))
 		return;
 	
-	g_flPlayerSpeedMultiplier[iClient] /= ICE_SLOWDOWN;
-	
-	//Recalculate player's speed
-	TF2_AddCondition(iClient, TFCond_SpeedBuffAlly, 0.01);
+	TF2_StunPlayer(iClient, 0.0, 0.0, TF_STUNFLAG_SLOWDOWN);
```
FortyTwoFortyTwo: `Timer_ResetSpeed` callback can be deleted when there `TF2_StunPlayer` with `ICE_DURATION`? Otherwise removing `TFCond_Dazed` cond also works.

PR #72, opened 2019-11-27
Refactor header

FortyTwoFortyTwo:
- Always use `TFTeam` enum instead of `int` for team variables
- Replace `BOSS_TEAM` and `ATTACK_TEAM` to new `const TFTeam`, `TFTeam_Boss` and `TFTeam_Attack`
- Remove some redundant `vsh`/`hale` prefix, we all know entire plugin is for VSH anyway
- Remove many unused attrib defines
- Other tiny bits im lazy to list


FortyTwoFortyTwo:
Just found out i accidentally removed engi's building extra stats while healed from medigun in PR #27, which is why build passes while removing `ATTRIB_SENTRYATTACKSPEED` and `ATTRIB_BIDERECTIONAL` attribs. Need to readd that, but not in this PR

PR #75, opened 2019-11-29
Add custom yeti model

Mikusch:
Since bone merging causes the model to be invisible sometimes, we will have to use a custom model.

This PR adds a custom yeti model and suppresses a few things related to the heavy's high noon kill taunt due to the model having a custom taunt sequence.


materials/models/player/boss/yeti/yeti_face_invun.vmt
```diff
@@ -0,0 +1,128 @@
+// Modified to use heavy's stuff by Kirillian,, custom invulnerability textures by Kirillian.
```
FortyTwoFortyTwo: double `,` after `Kirillian`. Also calling name twice? I expect words to be along something like this:
```
Modified to use as heavy's custom model, and custom invulnerability textures made by Kirillian.
```

Also adding name and link makes me think others need to ask for permission, which that is not true as Kirillan is fine with it. Worth adding that everyone is allowed to use without permission

materials/models/player/boss/yeti/yeti_face_invun.vmt
```diff
@@ -0,0 +1,128 @@
...
+			"resultVar" "$selfillumfresnelminmaxexp[1]"
+		}
+		"Equals"
+		{
+			"srcVar1" "$invulnphong"
+			"resultVar" "$phongboost" 
+		}
+
+
+
```
FortyTwoFortyTwo: 🍉

materials/models/player/boss/yeti/yeti_invun.vmt
```diff
@@ -0,0 +1,128 @@
...
+			"resultVar" "$selfillumfresnelminmaxexp[1]"
+		}
+		"Equals"
+		{
+			"srcVar1" "$invulnphong"
+			"resultVar" "$phongboost" 
+		}
+
+
+
```
FortyTwoFortyTwo: 🍉

PR #82, opened 2019-12-08
Dragon Furry

StarInAPond:
repressurization rate fix


FortyTwoFortyTwo:
Attribute sadly did not help into fixing it, i believe valve hardcoded to weapon's classname. Feel free to create issue about this. I'll take a look more about this later, whenever if there a (simple) way to fix this or not.

PR #94, opened 2019-12-17
Post-post-playtest Über Ranger update

woisalreadytaken:
This is where the balance changes come in. Mini rangers seem to be too weak, and that's fair, I didn't take any time thinking about how much health they should have besides "being tankier than Seeldier minions." This update makes ranger companions more unique compared to the other types of summons, essentially actual minibosses.

**Changes:**
Companions:
- Instead of a flat 600 base health, they now get 400 hp + 40 per player (on the enemy team and alive)
- They now can super jump, with weaker strength (half of an average jump's height)
- They're übered until 3 seconds after they make their first move on spawn (more details on this later)
- They get an outline for a short amount of time after spawning so players can tell which one is the boss

Boss:
- 100% Rage now lasts for 5 seconds instead of 6, and 200% lasts for 8 seconds instead of 7.998

Going into the technical side, with summoned players being the most important part of a boss' ability, it's important that the boss doesn't get fucked by their (lack of) action, so AFK players will be replaced accordingly as follows:
- When a player spawns, they'll have a 6 second countdown until they start moving
    - Only directional keys are considered. Jumping/looking/attacking/etc in place will be ignored
    - The moment they start moving, it goes away
    - If it ends and they haven't moved yet, the next best miniboss candidate will take their place
    - If there are no other candidates, that's very unfortunate. The spawn protection will go away but the AFK player won't go anywhere

Note that this just aims to catch legit AFK players without punishment, if people are purposefully suiciding or similar, usage of the punishment feature should be considered.

Known issue:
- The super jump text in the hud doesn't go away when you die, though this is a general issue rather than this boss's and should be done in a separate PR.

### DIFF
--- a/addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
@@ -55,8 +55,8 @@ methodmap CUberRanger < SaxtonHaleBase
 		abilityJump.iJumpChargeBuild = 1;
 		
 		CRageAddCond rageCond = boss.CallFunction("CreateAbility", "CRageAddCond");
-		rageCond.flRageCondDuration = 6.0;
-		rageCond.flRageCondSuperRageMultiplier = 1.333;	//This is 7.998 seconds, close enough
+		rageCond.flRageCondDuration = 5.0;
+		rageCond.flRageCondSuperRageMultiplier = 1.6;	// 8 seconds
 		rageCond.AddCond(TFCond_UberchargedCanteen);
 		
 		boss.iBaseHealth = 650;
@@ -68,7 +68,7 @@ methodmap CUberRanger < SaxtonHaleBase
 		for (int i = 1; i <= MaxClients; i++)
 			g_bUberRangerPlayerWasSummoned[i] = false;
 			
-		UberRangerResetColorList();
+		UberRanger_ResetColorList();
 	}
 	
 	public void GetBossName(char[] sName, int length)
@@ -85,7 +85,7 @@ methodmap CUberRanger < SaxtonHaleBase
 		StrCat(sInfo, length, "\n- Equipped with a Medi Gun");
 		StrCat(sInfo, length, "\n ");
 		StrCat(sInfo, length, "\nRage");
-		StrCat(sInfo, length, "\n- Übercharge for 6 seconds");
+		StrCat(sInfo, length, "\n- Übercharge for 5 seconds");
 		StrCat(sInfo, length, "\n- Summons a fellow Über Ranger");
 		StrCat(sInfo, length, "\n- Über Rangers are allowed to heal and über each other");
 		StrCat(sInfo, length, "\n- 200%% Rage: extends über duration to 8 seconds and summons 3 Über Rangers");
@@ -198,8 +198,9 @@ methodmap CUberRanger < SaxtonHaleBase
 				
 				boss.CallFunction("CreateBoss", "CMinionRanger");
 				TF2_RespawnPlayer(iClient);
-				
 				TF2_TeleportToClient(iClient, this.iClient);
+				
+				
 				TF2_AddCondition(iClient, TFCond_Ubercharged, 2.0);
 				
 				aValidMinions.Erase(iBestClientIndex);
@@ -282,8 +283,12 @@ methodmap CMinionRanger < SaxtonHaleBase
 {
 	public CMinionRanger(CMinionRanger boss)
 	{
-		boss.iBaseHealth = 600;
-		boss.iHealthPerPlayer = 0;
+		CBraveJump abilityJump = boss.CallFunction("CreateAbility", "CBraveJump");
+		abilityJump.iJumpChargeBuild = 1;
+		abilityJump.flMaxHeigth = 650.0;
+		
+		boss.iBaseHealth = 400;
+		boss.iHealthPerPlayer = 40;
 		boss.nClass = TFClass_Medic;
 		boss.iMaxRageDamage = -1;
 		boss.flWeighDownTimer = -1.0;
@@ -330,7 +335,7 @@ methodmap CMinionRanger < SaxtonHaleBase
 		
 		//Checking if the list is there at all or has been emptied
 		if (g_aUberRangerColorList == null || g_aUberRangerColorList.Length <= 0)
-			UberRangerResetColorList();
+			UberRanger_ResetColorList();
 			
 		//Assign color
 		int iColor[4];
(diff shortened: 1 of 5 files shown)


addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -402,8 +432,96 @@ public void UberRangerResetColorList()
...
+		return;
+	}
+	
+	//Snap the AFK player. There's no point in killing them if they're the only acceptable client available
+	ArrayList aValidMinions = new ArrayList();
+	GetValidSummonableClients(aValidMinions);
+	int iLength = aValidMinions.Length;
+	bool bPlayerAvailable = false;
+	
+	if (iLength > 0)
```
FortyTwoFortyTwo: Unneeded check when `for` won't bother loop in this case

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -278,19 +244,31 @@ methodmap CUberRanger < SaxtonHaleBase
 	}
 };
 
+static bool g_bUberRangerMinionHasMoved[TF_MAXPLAYERS+1];
+static Handle g_hUberRangerMinionAFKTimer[TF_MAXPLAYERS+1];
+static int g_iUberRangerMinionAFKTimeLeft[TF_MAXPLAYERS+1];
```
FortyTwoFortyTwo: Prefer all of those next to `g_bUberRangerPlayerWasSummoned`

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -360,16 +343,64 @@ methodmap CMinionRanger < SaxtonHaleBase
 			SetEntProp(iWearable, Prop_Send, "m_nModelIndexOverrides", g_iUberRangerBlightedBeak);
 			SetEntityRenderColor(iWearable, iColor[0], iColor[1], iColor[2], iColor[3]);
 		}
+		
+		g_hUberRangerMinionAFKTimer[this.iClient] = CreateTimer(0.0, Timer_UberRanger_ReplaceMinion, this.iClient);
```
FortyTwoFortyTwo: `g_hUberRangerMinionAFKTimer` should be set to null from `Destory` and/or `OnDeath`

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -402,8 +432,96 @@ public void UberRangerResetColorList()
...
+
+		boss.CallFunction("CreateBoss", "CMinionRanger");
+		TF2_RespawnPlayer(iBestClientIndex);
+
+		//Duration of this condition will reset when they move
+		TF2_AddCondition(iBestClientIndex, TFCond_UberchargedCanteen, 7.0);
+	}
+	
+	//Returns -1 if it finds nobody suitable
+	return iBestClientIndex;
```
FortyTwoFortyTwo: `return -1;`

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -402,8 +432,96 @@ public void UberRangerResetColorList()
...
+	bool bPlayerAvailable = false;
+	
+	if (iLength > 0)
+	{
+		for (int i = 0; i < iLength; i++)
+		{
+			int iCandidate = aValidMinions.Get(i);
+			if (!g_bUberRangerPlayerWasSummoned[iCandidate])
+			{
+				bPlayerAvailable = true;
```
FortyTwoFortyTwo: Can kill player here without needing `bPlayerAvailable`

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -402,8 +432,96 @@ public void UberRangerResetColorList()
...
+	if (g_iUberRangerMinionAFKTimeLeft[iClient] > 0)
+	{
+		g_iUberRangerMinionAFKTimeLeft[iClient]--;
+		g_hUberRangerMinionAFKTimer[iClient] = CreateTimer(1.0, Timer_UberRanger_ReplaceMinion, iClient);
+		return;
+	}
+	
+	//Snap the AFK player. There's no point in killing them if they're the only acceptable client available
+	ArrayList aValidMinions = new ArrayList();
+	GetValidSummonableClients(aValidMinions);
```
FortyTwoFortyTwo: Why not create new `ArrayList` inside `GetValidSummonableClients` without needing to pass handle to param

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
woisalreadytaken: this isn't meant to always return -1, I changed the comment to make it clearer

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -402,8 +432,96 @@ public void UberRangerResetColorList()
...
+
+		//Allow them to join the boss team
+		Client_AddFlag(iBestClientIndex, ClientFlags_BossTeam);
+		TF2_ForceTeamJoin(iBestClientIndex, TFTeam_Boss);
+
+		boss.CallFunction("CreateBoss", "CMinionRanger");
+		TF2_RespawnPlayer(iBestClientIndex);
+
+		//Duration of this condition will reset when they move
+		TF2_AddCondition(iBestClientIndex, TFCond_UberchargedCanteen, 7.0);
```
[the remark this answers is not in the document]
woisalreadytaken: it's not infinite in the (rare) case it runs out of acceptable candidates and leaves the guy there

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -330,15 +308,20 @@ methodmap CMinionRanger < SaxtonHaleBase
...
+		g_aUberRangerColorList.Erase(0);
 		iColor[3] = 255;
 		
 		SetEntityRenderColor(this.iClient, iColor[0], iColor[1], iColor[2], iColor[3]);
-		g_aUberRangerColorList.Erase(0);
+		
+		//Add glow for him to be easily recognizable as not the boss during spawn uber
+		//Round started check is there so it doesn't show up when spawning on the next round as well
+		if (g_bRoundStarted)
+			CreateTimer(3.0, Timer_EntityCleanup, TF2_CreateGlow(this.iClient, iColor));
```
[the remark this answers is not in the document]
woisalreadytaken: the glow has the colors they get on spawn, so it makes sense for it to be there

addons/sourcemod/scripting/vsh/bosses/boss_uberranger.sp
```diff
@@ -282,15 +251,23 @@ methodmap CMinionRanger < SaxtonHaleBase
 {
 	public CMinionRanger(CMinionRanger boss)
 	{
-		boss.iBaseHealth = 600;
-		boss.iHealthPerPlayer = 0;
+		CBraveJump abilityJump = boss.CallFunction("CreateAbility", "CBraveJump");
+		abilityJump.iJumpChargeBuild /= 4;						//4x slower jump charge rate
+		abilityJump.flMaxHeigth /= 2;							//Half max height for super jumps
```
FortyTwoFortyTwo: Huge tabs

PR #97, opened 2019-12-18
Change certain on-screen text behavior

woisalreadytaken:
Just minor changes regarding HUD text, here's a list of them:
- By default, text will now go away on death. This was most commonly seen in duo bosses (and would probably be noticed the most with minibosses in the future), includes text like rage meter, super jump info, boss/ability info, etc. and does not include important text such as boss health, your own damage/assistance and damage/assistance from people who you are spectating.
- (Multi-)Bosses who die will now be able to see remaining boss(es)' health and damage/assistance from people who are being spectated.
- Added info text for the wall climbing ability, as it's not obvious to unfamiliar people.


woisalreadytaken:
I don't see any harm in letting dead bosses see people's damage

PR #104, opened 2020-01-05
Update dome to only damage one team based from CP owner

### LINKED ISSUE #99: Team Dome
Lots of ideas by Zane so thanks for that.

Currently dome deals damage to both team, pushing both to near death. However usually only one team need to be in penalty from dome, depending on which team stalls. So this Team Dome idea popped up for this.

- Dome does not appear/start automatically, But CP unlocks instead. likely using same timer as current dome.
- Capture time should not be based from map's time, always same/hard set from plugin.
- Both team can capture. Team cannot block capture progress, but capture rate will be slowed down instead.
- Once CP is captured, round won't end but dome will spawn instead. Dome color will be same as whoever team captured point.
- Dome will only deal damage to whoever team not owner of CP capped.
- Team can recapture CP to flip dome color and damage to opposite team.
- Dome speed will likely be faster to compensate with cap required to start dome.
- Dome freeze depends on who captured CP. If attack team, freeze when attack team died. If boss team, freeze when attack team damages boss. This idea is for losing team to counter back captured team, showing less stall happening.
- Somehow deal with map's new dome pos from config, possibly X radius needing to be nearby CP to capture.

FortyTwoFortyTwo:
closes #99 

Decided to start this somewhat simple from what was originally planned:
- On round start, dome unlocks itself in `60 + (5 * N)` seconds where n is amount of attack players
- Takes 15 seconds to capture point for 1 player
- Boss have a 2x capture rate (unlike usual 3x rate)
- Dome takes 2 mins to fully shrink from start. Fixed speed, but depends on start radius on many maps.
- Dome will never slow down

Many of those numbers can be can be changed in config.

TODO
- Increase dome damages?
- Make capturing CP works with `vsh_dome_centre` convar, CP pos moved.

Other than that, rest is pretty much from what said in #99

### DIFF
--- a/addons/sourcemod/configs/vsh/maps.cfg
+++ b/addons/sourcemod/configs/vsh/maps.cfg
@@ -221,10 +221,6 @@
 		{
 			"vsh_dome_radius_start"	"2500"
 		}
-//		"vsh_streets"	//This map really only used for plugin testing purposes
-//		{
-//			"vsh_dome_enable"		"0"
-//		}
 		"vsh_teul"
 		{
 			"vsh_dome_radius_start"	"2250"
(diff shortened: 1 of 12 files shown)


FortyTwoFortyTwo:
This is ready to go

PR #117, opened 2020-01-09
Fix Blutsauger not self-healing

woisalreadytaken:
The Blutsauger lost its ability to self-heal in the tags rework.


woisalreadytaken:
I thought of doing that but I also thought this would be much more simple but I ended up doing that anyway

added an `"overheal"` parameter to both `Tags_AddHealth` and `Tags_AddHealthBase`, which is the multiplier of the target's max health that the (over)healing will cap at, defaulting at 1.5

PR #118, opened 2020-01-11
Fix crash from electric modifier endless loop

FortyTwoFortyTwo:
https://crash.limetech.org/yppcfvatnaip

I suspect this started from PR #106, `SDKHooks_TakeDamage` native bypassing `SDKHook_OnTakeDamage` hook, but apparently not `SDKHook_OnTakeDamageAlive`. This causes an endless loop of `SDKHooks_TakeDamage` calling `OnAttackDamage` boss function.

### DIFF
--- a/addons/sourcemod/scripting/vsh/modifiers/modifiers_electric.sp
+++ b/addons/sourcemod/scripting/vsh/modifiers/modifiers_electric.sp
@@ -8,11 +8,6 @@ methodmap CModifiersElectric < SaxtonHaleBase
 	{
 	}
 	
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo:
Nevermind the fix does not fully help, giving a different crash log instead
https://crash.limetech.org/irsoiq3c73lt

FortyTwoFortyTwo:
Crash log i linked is likely just vrlab only thing and not related to this. Also fixed crash seemly common for windows local servers.

PR #127, opened 2020-01-17
Change Pain Train to suit the team dome mechanic

woisalreadytaken:
Now that there's a control point to capture, the stock upside of the Pain Train shouldn't be touched, and this PR reverts it.
The +10% bullet vulnerability stat is changed to +10% general damage vulnerability so the weapon has an actual, yet simple, downside.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -529,8 +529,8 @@
 		
 		"154"	//Pain Train
 		{
-			"desp"			"Pain Train: {positive}+30% health from healers, {negative}-50% health from health packs"
-			"attrib"		"70 ; 1.3 ; 109 ; 0.5 ; 68 ; 0.0 ; 67 ; 1.0"
+			"desp"			"Pain Train: {negative}+10% damage vulnerability from all sources"
+			"attrib"		"67 ; 1.0 ; 412 ; 1.1"
 		}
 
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -529,8 +529,8 @@
 		
 		"154"	//Pain Train
 		{
-			"desp"			"Pain Train: {positive}+30% health from healers, {negative}-50% health from health packs"
-			"attrib"		"70 ; 1.3 ; 109 ; 0.5 ; 68 ; 0.0 ; 67 ; 1.0"
+			"desp"			"Pain Train: {negative}+10% damage vulnerability from all sources"
```
Mikusch: Does the ``from all sources`` really need to be specified explicitly? Attribute description itself is ``%s1% damage vulnerability on wearer``, should be clear enough imo

woisalreadytaken: I felt like it needed to be specified because the weapon already has some kind of damage vulnerability, so telling the player it moves from x to y type of vulnerability should be more important than leaving it at just "damage," possibly making someone wonder "doesn't it already have that stat"

FortyTwoFortyTwo: `from all sources` does sound better when weapon already have bullet vulnerability

PR #137, opened 2020-01-23
Baby Face's Blaster: increase damage req to 400

woisalreadytaken:
In VSH, this weapon's charge meter is much easier to build as the threats for its downside simply do not exist, making the default speed gain feel annoying and undeserved.

Instead of trying to reinvent the wheel, the speed boost will remain a feature of the weapon, it'll just be a bit harder to get. :)

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -546,8 +546,8 @@
 
 		"772"	//Baby Face Blaster
 		{
-			"desp"			"Baby Face Blaster: {negative}Boost takes 400 damage to fully charge and is reset on stun"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -546,7 +546,8 @@
 
 		"772"	//Baby Face Blaster
 		{
-			"desp"			"Baby Face Blaster: {negative}Boost is reset on stun"
+			"desp"			"Baby Face Blaster: {negative}Boost takes 400 damage to fully charge and is reset on stun"
```
FortyTwoFortyTwo: iirc last time i tested, Soda Popper actually takes 380 damage to fill up meter (valve why)

PR #142, opened 2020-02-02
Allow Vagineer to have more than one sentry up on rage

woisalreadytaken:
Right now, if Vagineer rages while already having an active sentry, it'll be destroyed so another one can be built. That's pretty harsh for the boss as his rage doesn't have much going on defensively besides a little-ranged stun and with this happening commonly in a panic-raging scenario, he's usually getting shredded.

This PR makes it so the previous sentry won't be destroyed while allowing ownership of multiple sentries.

### DIFF
--- a/addons/sourcemod/scripting/vsh/bosses/boss_vagineer.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_vagineer.sp
@@ -187,6 +187,7 @@ methodmap CVagineer < SaxtonHaleBase
 		
 		SetVariantInt(iSentryHealth);
 		AcceptEntityInput(iEntity, "SetHealth");	//Sets sentry health
+		SDK_RemoveObject(this.iClient, iEntity);	//Make the boss not the sentry's original owner so he gets to potentially build more of them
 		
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/vsh/bosses/boss_vagineer.sp
```diff
@@ -187,6 +187,7 @@ methodmap CVagineer < SaxtonHaleBase
 		
 		SetVariantInt(iSentryHealth);
 		AcceptEntityInput(iEntity, "SetHealth");	//Sets sentry health
+		SDK_RemoveObject(this.iClient, iEntity);	//Make the boss not the sentry's original owner so he gets to potentially build more of them
```
FortyTwoFortyTwo: Do we want to remove/destroy sentry when Vagineer class get `Destroy` mid-game? `RemoveObject` leaves sentry alive when engineer changes class/team, and not decaying health as vagineer is destroyed

woisalreadytaken: ok

PR #144, opened 2020-02-04
Nuke caber

woisalreadytaken:
This change focuses on what the caber is supposed to be: a weapon to use when you have no other options.

For the simple part, all this really does is amplify what the caber already does: you deal more damage with it and never survive damage from it (and by never I really mean it), while adding fitting visual and sound effects to it. Explosion radius is unchanged.

A couple of things for the config were also changed to support this and maybe other things in the future:
- **Tags_Explode** now has custom `particle` and `sound` parameters, if the tag is used but these parameters aren't specified it'll use the same visual/sound effects it always has.
    - The `sound` parameter supports up to 16 non-custom sounds and will play one of them at random for the explosion.
- **Tags_ForceSuicide** has been added, it just kills the target.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1248,7 +1248,7 @@
 					"target"		"melee"
 					"type"			"float"
 					"prop"			"m_flModelScale"
-					"value"			"3.0"
+					"value"			"2.0"
 				}
 			}
 			
@@ -1262,6 +1262,15 @@
 				"params"
 				{
 					"multiply"		"3.0"	//Multiply explosion damage by 3
+				}
+			}
+			
+			"takedamage"
+			{
+				"filter"
+				{
+					"attackweapon"	"melee"
+					"damagetype"	"blast"	//Only when hit by melee blast damage
 				}
 				
 				"Tags_Explode"
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -155,8 +155,12 @@
 //Tags_Explode            - Creates explosion at target
 //- damage                  - Explosion damage
 //- radius                  - Radius of explosion
+//- particle                - Optional particle effect. Set to a generic explosion particle by default
+//- sound                   - Optional sound of the explosion, up to 16 sounds can be specified. Set to air strike explosion sounds by default. Does not support custom sounds
 //
 //Tags_KillWeapon         - Unequips and destroy weapon target
+//
+//Tags_ForceSuicide       - Forces the target to suicide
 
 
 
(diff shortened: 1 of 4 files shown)


addons/sourcemod/scripting/saxtonhale.sp
```diff
@@ -32,6 +32,7 @@
 #define MAX_BUTTONS 					26
 #define MAX_TYPE_CHAR					32		//Max char size of methodmaps name
 #define MAXLEN_CONFIG_VALUE 			256		//Max config string buffer size
+#define MAXLEN_CONFIG_ALLVALUES			1024	//Max config buffer size for groups of values
```
FortyTwoFortyTwo: `MAXLEN_CONFIG_VALUEARRAY`? May fit better as it should only be used on possible arrays

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1234,20 +1238,54 @@
...
+					"damagetype"	"blast"	//Only when the caber detonates
 				}
 				
 				"params"
 				{
-					"multiply"		"2.0"	//Multiply explosion damage by 2
+					"multiply"		"3.0"	//Multiply explosion damage by 3
+				}
+				
+				"Tags_Explode"
```
FortyTwoFortyTwo: Should add target param to victim

addons/sourcemod/scripting/vsh/tags/tags_params.sp
```diff
@@ -50,11 +50,11 @@ methodmap TagsParams < StringMap
 	{
 		if (this == null) return false;
 		
-		char sValue[MAXLEN_CONFIG_VALUE];
+		char sValue[MAXLEN_CONFIG_ALLVALUES];
 		if (!this.GetString(sKey, sValue, sizeof(sValue)))
 			return false;
 		
-		char sValues[32][32];
+		char sValues[16][MAXLEN_CONFIG_VALUE];
```
FortyTwoFortyTwo: May well then add new define for `16`, `MAX_CONFIG_ARRAY`?

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1234,20 +1238,54 @@
...
+			"attrib"		"414 ; 1.0 ; 207 ; 11.0"
+			
+			"spawn"
+			{
+				"Tags_SetEntProp"	//Make the caber fucking massive
+				{
+					"target"		"melee"
+					"type"			"float"
+					"prop"			"m_flModelScale"
+					"value"			"3.0"
```
FortyTwoFortyTwo: I dunno, 3.0 may be a bit overkill but it just a visual change

addons/sourcemod/scripting/vsh/tags.sp
```diff
@@ -783,11 +783,19 @@ public void Tags_Explode(int iClient, int iTarget, TagsParams tParams)
...
 	
+	//If no particle was specified, pick a generic explosion particle
+	char sParticle[MAXLEN_CONFIG_VALUE];
+	if (!tParams.GetString("particle", sParticle, sizeof(sParticle)))
+		Format(sParticle, sizeof(sParticle), "ExplosionCore_MidAir");
+	
+	//If no sounds were specified, pick a generic explosion sound. If multiple sounds were specified, pick a random one
+	char sSound[MAXLEN_CONFIG_VALUE];
+	if (!tParams.GetStringRandom("sound", sSound, sizeof(sSound)))
+		Format(sSound, sizeof(sSound), "weapons/airstrike_small_explosion_0%i.wav", GetRandomInt(1,3));
```
FortyTwoFortyTwo: `%d` and spacing between `1,` and `3` random numbers

addons/sourcemod/configs/vsh/vsh.cfg
woisalreadytaken: did this in a different way, made the filter consider melee and blast damage instead

PR #150, opened 2020-03-07
Fix the airblast visual bug

haxtonsale:
`m_iWeaponState` remained at 3 (FlamethrowerState_Airblast) if pyro airblasted while he was holding M1. This caused airblast particles to appear with flame particles for everyone but the pyro, as long as he was holding M1.


haxtonsale:
Idk but since we need to know whether `m_iWeaponState` needs to be set to `FlamethrowerState_Firing` or `FlamethrowerState_Idle` doing it in `Tags_OnButton` seems to be appropriate.

PR #153, opened 2020-03-11
Pyrocar fix

Alex-Turtle:
Added the missing attributes and added a speedboost to the float ability to give the boss a chance to catch scouts


addons/sourcemod/scripting/vsh/abilities/ability_rage_hop.sp
```diff
@@ -118,15 +121,15 @@ methodmap CRageHop < SaxtonHaleBase
...
+				TF2_Explode(this.iClient, vecExplosionOrigin, this.flBombDamage * 1.5, this.flBombRadius * 1.25, "heavy_ring_of_fire", sSound);
 			else
 				TF2_Explode(this.iClient, vecExplosionOrigin, this.flBombDamage, this.flBombRadius, "heavy_ring_of_fire", sSound);
 			
 			for (int i = 1; i <= MaxClients; i++)
 			{
 				if (IsClientInGame(i) && IsPlayerAlive(i) && GetClientTeam(i) > 1 && GetClientTeam(i) != iTeam)
 				{
-					if (IsClientInRange(i, vecExplosionOrigin, 200.0))
+					if (IsClientInRange(i, vecExplosionOrigin, this.flBombRadius) || (IsClientInRange(i, vecExplosionOrigin, this.flBombRadius * 1.25) && this.bSuperRage))
```
FortyTwoFortyTwo: I would be more happy if radius is calculated to local variable and test 1 `IsClientInRange`, rather than having 2 `IsClientInRange`

addons/sourcemod/scripting/vsh/bosses/boss_pyrocar.sp
```diff
@@ -85,10 +85,9 @@ methodmap CPyroCar < SaxtonHaleBase
 	{
 		StrCat(sInfo, length, "\nHealth: Low");
 		StrCat(sInfo, length, "\nYou are forced to go forward");
-		StrCat(sInfo, length, "\nYou have the same speed as the medic");
```
FortyTwoFortyTwo: I think it worth to keep comment that it slower than other bosses

PR #166, opened 2020-05-06
decrease chance of becoming default hale

tacticturt:
There are 11 other solo bosses (excluding The Last Yeti and Merasmus) with, to be frank, more interesting kits than the default hale. While he may be the "face" of VSH, there are many players who enjoy playing as/against the "special" hales instead, and the smaller amount who are tired of playing as/against him.

If my math is right, he should now have just under double the chance of other bosses to picked, rather than the more-than-triple chance he has currently:

New formula: 100 - 15% chance of being hale = 85% chance for other bosses (excluding duo-bosses.) 85 / 11 = 7.7% chance for each boss other than hale.

Old formula: 100 - 25% chance of being hale = 75% chance for other bosses (excluding duo-bosses) 75/ 11 = 6.8% chance for each boss other than hale.

### DIFF
--- a/addons/sourcemod/scripting/vsh/nextboss.sp
+++ b/addons/sourcemod/scripting/vsh/nextboss.sp
@@ -7,7 +7,7 @@ void NextBoss_Init()
 	g_aNextBoss = new ArrayList(sizeof(NextBoss));
 	g_aNextBossMulti = new ArrayList();
 	
-	g_ConfigConvar.Create("vsh_boss_chance_saxton", "0.25", "% chance for next boss to be Saxton Hale from normal bosses pool (0.0 - 1.0)", _, true, 0.0, true, 1.0);
+	g_ConfigConvar.Create("vsh_boss_chance_saxton", "0.15", "% chance for next boss to be Saxton Hale from normal bosses pool (0.0 - 1.0)", _, true, 0.0, true, 1.0);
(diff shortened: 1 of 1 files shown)


Mikusch:
I don't really agree with this. While he isn't as interesting as the other bosses, Saxton Hale *should* be much more common. Current percentage is fine IMO.

Blazephlozard:
Personally I would rather have a higher chance of Saxton than a lower chance...

Perhaps a "Saxton Preference" option could be implemented. Off = equal chance for all bosses, On = the percentage it is now

FortyTwoFortyTwo:
I thought current chance for hale to appear were very high, but actually not as high as i thought of.
Just under twice the chance for hale does seem a bit overkill, 20% chance would make it 2.75 times more which still feels somewhat low.
I'll leave this to others to give more thoughts on it.

> Perhaps a "Saxton Preference" option could be implemented. Off = equal chance for all bosses, On = the percentage it is now

Would slightly alter gameplay and chances of winning which i tends to avoid that

tacticturt:
> Just under twice the chance for hale does seem a bit overkill, 20% chance would make it 2.75 times more which still feels somewhat low.

The number can always be changed to something higher than 15%, like 20% as you said.

FortyTwoFortyTwo:
Marjory of people doesn't seems to want to have chances changed, so will leave it as it is

PR #168, opened 2020-05-11
Pyrocar rework

Alex-Turtle:
Changelog:

-Slightly faster movement speed (360 instead of 345)
-100% increased damage plus pierces resistances
-Halved Flamethrower range
-Afterburn duration reduced by 50%, and damage reduced by 75%
-Better float ability: You don't need to fully charge it to activate it plus you go upwards faster
-Rage rework, explosive damage is now scaled by the falling speed instead of a constant value. Rage now lasts 8 and 12 seconds for super rage respectively (from 10 and 15)


addons/sourcemod/scripting/vsh/abilities/ability_force_forward.sp
```diff
@@ -96,4 +124,28 @@ methodmap CForceForward < SaxtonHaleBase
...
+		{
+			SDKHook(iEntity, SDKHook_StartTouch, OnCatapultStart);
+			SDKHook(iEntity, SDKHook_EndTouch, OnCatapultEnd);
+		}
+	}
+};
+
+public Action OnCatapultStart(int iEntity, int iClient)
+{
+	g_bInCatapult[iClient] = true;
```
FortyTwoFortyTwo, line 140: Another one, i believe `iClient`, the toucher is not always client, can be any other entity so a check would be needed

PR #169, opened 2020-05-14
Building's health drain

StarInAPond:
Small nerf to a strong rage


FortyTwoFortyTwo:
Don't think it needs a nerf, already hard to fight sentry in some cases

Issue #191, opened 2020-08-10
Boss Concept: Samyro


Waifers:
Another Pyro boss that isn't El Muchacho. Samyro is a more vanilla-like boss intended to co-exist with Pyrocar that utilizes an underused feature, Mannpowers, to create a rushdown-type boss.

What is Samyro: A chaotic Blu Pyro that uses a Half-Zatochi and wears The Last Breath and The Flamboyrant Flamenco. People who want more information can go to the wiki page below.
https://tf2freakshow.fandom.com/wiki/Samyro

Theme: https://www.youtube.com/watch?v=qae4OZZKsR4

Basic Statistics
- Mediocre base health (Comparable to HHH Jr.)
- Average speed
- Average damage requirement for rage 
- Uses the Half-Zatochi with the typical Hale Stats. 
- Can wall-climb

Passive: Endless Rage
- Samyro passively builds up additional rage over time.

M2: Agility 
- At 100% charge, Samyro can use this ability and gains the Agility Mannpower for a short period of time (possibly ~8 seconds).
- Large cooldown- Atleast 30 seconds. 
- Automatically put on cooldown if rage is activated for the duration of it.
(The bonuses the boss uses from the Agility Mannpower are the large movespeed increase and the jump height increase)

Rage: Applies the **Haste** Mannpower and the Battalion Buff for 12 seconds. 
- Overrides the Agility Mannpower
(The bonuses the boss uses from the Haste Mannpower are the movespeed increase and the firerate increase.)

200% Rage: +2 seconds of rage duration, additionally adds Uber during the duration.

This is the initial idea for the boss and his stats, right now it's open for any changes and recommendations.

PR #192, opened 2020-08-10
Samyro

Mikusch:
Implementation of #191 

There is still one issue with this: The boss doesn't actually gain the speed benefits of the two applied powerup conditions due to the way VSH sets the boss speed. Will probably change that later, either in this or a separate PR. Once that is done, this boss should be balanced quite nicely.

### DIFF
--- a/addons/sourcemod/scripting/saxtonhale.sp
+++ b/addons/sourcemod/scripting/saxtonhale.sp
@@ -356,6 +356,7 @@ ConVar tf_arena_preround_time;
 #include "vsh/abilities/ability_brave_jump.sp"
 #include "vsh/abilities/ability_dash_jump.sp"
 #include "vsh/abilities/ability_drop_model.sp"
+#include "vsh/abilities/ability_conditions.sp"
 #include "vsh/abilities/ability_float_jump.sp"
 #include "vsh/abilities/ability_force_forward.sp"
 #include "vsh/abilities/ability_groundpound.sp"
@@ -390,6 +391,7 @@ ConVar tf_arena_preround_time;
 #include "vsh/bosses/boss_pyromancer_scorched.sp"
 #include "vsh/bosses/boss_pyromancer_scalded.sp"
 #include "vsh/bosses/boss_redmond.sp"
+#include "vsh/bosses/boss_samyro.sp"
 #include "vsh/bosses/boss_seeldier.sp"
 #include "vsh/bosses/boss_seeman.sp"
 #include "vsh/bosses/boss_uberranger.sp"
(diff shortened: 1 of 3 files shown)


addons/sourcemod/scripting/vsh/bosses/boss_samyro.sp
```diff
@@ -0,0 +1,219 @@
...
+};
+
+public Action Command_DropItem(int iClient, const char[] sCommand, int iArgs)
+{
+	//Prevent boss from dropping powerups
+	if (SaxtonHaleBase(iClient).bValid)
+		return Plugin_Handled;
+	
+	return Plugin_Continue;
+}
```
FortyTwoFortyTwo, line 219: May well then move this to `console.sp` if we gonna block `dropitem` to all bosses

addons/sourcemod/scripting/vsh/bosses/boss_samyro.sp
```diff
@@ -0,0 +1,219 @@
...
+
+methodmap CSamyro < SaxtonHaleBase
+{
+	public CSamyro(CSamyro boss)
+	{
+		boss.CallFunction("CreateAbility", "CWallClimb");
+		
+		CAddCond abilityCond = boss.CallFunction("CreateAbility", "CAddCond");
+		abilityCond.AddCond(TFCond_RuneAgility);
+		abilityCond.bRemoveOnRage = true;
```
FortyTwoFortyTwo, line 58: Seems like both mannpower conds can be gained if rage is used first, then m2 ability during rage

addons/sourcemod/scripting/vsh/bosses/boss_samyro.sp
```diff
@@ -0,0 +1,219 @@
...
+	{
+		//Übercharge on 200% rage
+		if (this.bSuperRage)
+		{
+			CRageAddCond rageCond = this.CallFunction("FindAbility", "CRageAddCond");
+			if (rageCond != INVALID_ABILITY)
+			{
+				TF2_AddCondition(this.iClient, TFCond_Ubercharged, rageCond.flRageCondDuration * rageCond.flRageCondSuperRageMultiplier);
+			}
+		}
```
FortyTwoFortyTwo, line 156: May be worth to add `CAddCond` support for super rage cond

addons/sourcemod/scripting/vsh/bosses/boss_samyro.sp
```diff
@@ -0,0 +1,219 @@
...
+	}
+	
+	public void GetBossInfo(char[] sInfo, int length)
+	{
+		StrCat(sInfo, length, "\nHealth: Low");
+		StrCat(sInfo, length, "\n ");
+		StrCat(sInfo, length, "\nAbilities");
+		StrCat(sInfo, length, "\n- Passive Rage Gain");
+		StrCat(sInfo, length, "\n- Wall Climb");
+		StrCat(sInfo, length, "\n- Agility");
```
FortyTwoFortyTwo, line 86: Doesnt really tell much about this, "Agility powerup on alt attack" or similar should be enough

addons/sourcemod/scripting/saxtonhale.sp
```diff
@@ -356,6 +356,7 @@ ConVar tf_arena_preround_time;
 #include "vsh/abilities/ability_brave_jump.sp"
 #include "vsh/abilities/ability_dash_jump.sp"
 #include "vsh/abilities/ability_drop_model.sp"
+#include "vsh/abilities/ability_conditions.sp"
```
FortyTwoFortyTwo, line 359: c before d in order

PR #200, opened 2021-02-12
Shortstop

StarInAPond:
Reload speed buff for spam weapon Shortstop meant to be, low damage with constant rate of fire and high accuracy

35% brings reload time to 0.98 seconds compared to 1.52 old

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -590,6 +590,12 @@
 			}
 		}
 		
+		"220"	//Shortstop
+		{
+			"desp"			"Shortstop: {positive}+35% faster reload time"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -590,6 +590,12 @@
 			}
 		}
 		
+		"220"	//Shortstop
+		{
+			"desp"			"Shortstop: {positive}+35% faster reload time"
+			"attrib"		"97 ; 1.35"
```
FortyTwoFortyTwo, line 596: Meant to be `0.65` instead of `1.35`, this attrib makes smaller number quicker to reload while bigger number longer to reload

FortyTwoFortyTwo (approved):

PR #201, opened 2021-02-12
Scout pistols

StarInAPond:
Pistols, aside from Winger, are rarely useful and overshadowed by other secondaries, mainly because you run out of ammo too quickly.

+67% results in 2 additional clips, that makes scout carry 60 ammo, 1 small ammopack restores full clip.


FortyTwoFortyTwo (approved):
Would've suggested `67%` instead of `66%` for more accurate rounding, but I'll still take it anyway

PR #202, opened 2021-02-13
Back Scatter

StarInAPond:
Crits whenever it would normally mini-crit, no minicrits from usual sources (Crit-a-cola, medic beam, last survivors);
given weapon downsides and hard condition to get additional damage, this change should make Back Scatter useful


FortyTwoFortyTwo (commented):
I should've been more clearer on what `"minicrit"	"0"` affects, it does prevent minicrit from last 3 players and related to `"minicrit" "1"` from all-class-slot, but it does NOT affect Crit-a-Cola & Medigun, meaning crit would be applied from those.

Medigun issue should be possible to prevent as it given by VSH plugin, but since Crit-a-Cola is from TF2 it would be a lot more harder to prevent that.

PR #203, opened 2021-02-13
Loose cannon

StarInAPond:
It's been a long time since Eyelander speed nerf, Loose cannon shouldn't be problematic anymore


FortyTwoFortyTwo (commented):
I'm still in fear with Loose Cannon without downside, it's not a problematic right now because it's nerfed, but perhaps too much. I would be okay if one of the attrib nerf is kept or values changed.

PR #217, opened 2021-04-26
Bonk! Atomic Punch

StarInAPond:
Taunt eats 1 second from already short duration, 12 should be plenty, with 18 seconds window to recharge after effect.
On the other hand, I'm not sure anyone wants to see tankier scouts, might be good to tweak it to 10 seconds


FortyTwoFortyTwo (commented):
I don't think this change would make significant effect, but I'm unsure if it even worth to change duration, 8 secs was duration of normal bonk.

StarInAPond:
50% is pretty significant change, would definitely make melee scout (lol) somewhat viable

PR #219, opened 2021-04-26
Fan O'War

StarInAPond:
I feel like rage drain is just not a good concept for scout, if he has to go into melee range, effect has to be worth it, right now it's basically a weaker bat with delayed damage in a form of rage, and very crippling fire rate. Rage drain itself isn't even that good! You'll make hale use 1 less rage in ideal conditions, how many hits do you need to make it 2? Ten? Twenty? How about 35+?

Mark for death is a strong effect, I'm not sure making it last half as long (8, previously 15) would still be enough, decided to keep slow fire rate.

Open to discussions

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -655,7 +655,7 @@
 		
 		"355"	//Fan O'War
 		{
-			"desp"			"Fan O'War: {neutral}Replaces mark-for-death into a rage drain, {negative}100% slower firing speed"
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo (commented):
VSH has always been avoiding boss potentially getting minicrit debuff (mark for death & jarate) as red players has a moment to really grind out boss health.

I'm still against it, but I also agree current weapon aren't great, my only suggestion is either different effect on hit, or just increase rage loss.

StarInAPond:
I actually wanted to suggest buff to bosses' health at the end of all of this, that would've opened more opportunities for buffs

PR #222, opened 2021-04-26
Air Strike

StarInAPond:
Another boring suggestion for another weapon in a weird spot.
I feel like making explosion radius bigger should help Air Strike with it's gimmick, but something tells me it will still be used as an alternative to Liberty Launcher

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -722,6 +722,7 @@
 		{
 			"desp"			"Air Strike: {positive}Every 200 damages gains 1 clip size"
 			"tags"			"damage_head_required ; 200"
+			"attrib"		"99 ; 1.2"
 			
 			"attackdamage"
 			{


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -722,6 +722,7 @@
 		{
 			"desp"			"Air Strike: {positive}Every 200 damages gains 1 clip size"
```
FortyTwoFortyTwo, line 723: Missing comment for whatever changes to be made

FortyTwoFortyTwo (commented):
Air Strike itself already have -10% explosion radius penalty, is just removing said penalty enough?

StarInAPond:
Ideally, I'd want it to have something else, can't think of anything

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -720,8 +720,9 @@
 		
 		"1104"	//Air Strike
 		{
-			"desp"			"Air Strike: {positive}Every 200 damages gains 1 clip size"
+			"desp"			"Air Strike: {positive}Every 200 damages gains 1 clip size, +30% explosion radius"
 			"tags"			"damage_head_required ; 200"
+			"attrib"		"99 ; 1.2"
```
FortyTwoFortyTwo, line 725: This is now more confusing, +20% bonus (99) was added while -10% penalty (100) still stays, making overall bonus either +8% or +10% depending on math.

I'm not sure whenever if most people would read this desp as -10% penalty is also part of the percentage or ignored.

StarInAPond: Indeed, it's confusing. 
Changed description to mention +20% bonus while +33% is being applied for rounding, +20% radius splash **compared** to default rocket launcher, and 30% compared to old Air strike

PR #224, opened 2021-04-26
The Panic Attack

StarInAPond:
New Panic Attack is fine, old version just doesn't work well with vsh, loading shells leaves you very vulnerable, there's no real benefit, it's barely faster than shooting manually

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -500,13 +500,6 @@
 	{
 		// MULTI CLASS
 
-		"1153"	//Panic Attack
-		{
-			"desp"			"Panic Attack: {positive}+30% faster firing speed, {negative}hold fire required to load shells"
-			"attrib"		"6 ; 0.7 ; 710 ; 1.0"
-			"clip"			"0"
-		}
-		
 		"357"	//Half-Zatoichi
 		{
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo (commented):
I made changes to Panic Attack because of how identical Shotgun is.

This weapon could have different stats, if current vsh stats were to stay I would've just suggest increasing fire rate very high.

StarInAPond:
Maybe add something else to the new one?
Old just doesn't work due to spread, slow load and slow fire rate, you also can't change weapons while shooting, purely short range shotgun

Mikusch:
> I made changes to Panic Attack because of how identical Shotgun is.

It has different stats, I wouldn't call that identical. The weapon doesn't need to be different than it is in the regular game, there is absolutely no point in that.

As it is right now, it's pretty much unusable. Too slow and sluggish, basically just dead weight in your loadout.

FortyTwoFortyTwo:
> The weapon doesn't need to be different than it is in the regular game, there is absolutely no point in that.
A lot of VSH weapons already have some changes that some may seems not worth to change anyway. For this weapon, I'm okay with just VSH changes removed, but I just think this could also be the moment to have new different stats instead.

I don't think any of us have solid idea on what new stats could be given, it can be anything, or I can just take the normal stats

FortyTwoFortyTwo (approved):

PR #226, opened 2021-04-26
The Degreaser

StarInAPond:
The Degreaser is outclasses by 2 other Flamethrowers with airblast, and because main source of damage is afterburn, combos aren't very beneficial.
Damage in exchange of  knockback should fill Degreaser's utility role

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -922,6 +922,23 @@
 			}
 		}
 		
+		"215"	//The Degreaser
+		{
+			"desp"			"The Degreaser: {positive}Requires 300 damage for airblast"
+	
+				"spawn"
+				{
+					"Tags_Airblast"
+					{
+						"override"		"pyro-primary-airblast"
+						"target"		"primary"
+						"start"			"300"
+						"damage"		"300"	
+
+					}
+				}
+		}
+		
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -922,6 +922,23 @@
...
+				{
+					"Tags_Airblast"
+					{
+						"override"		"pyro-primary-airblast"
+						"target"		"primary"
+						"start"			"300"
+						"damage"		"300"	
+
+					}
+				}
```
FortyTwoFortyTwo, line 939: Extra tab makes this styling look weird

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -922,6 +922,23 @@
...
+		"215"	//The Degreaser
+		{
+			"desp"			"The Degreaser: {positive}Requires 300 damage for airblast"
+	
+				"spawn"
+				{
+					"Tags_Airblast"
+					{
+						"override"		"pyro-primary-airblast"
+						"target"		"primary"
```
FortyTwoFortyTwo, line 934: Don't think this line is needed as `target` is already set, unchanged from primary pyro

FortyTwoFortyTwo (approved):

PR #227, opened 2021-04-26
The Manmelter

StarInAPond:
This change doesn't make much sense, but I decided to go with it anyway, I feel like purely defensive role is the only one pyro's secondaries do not fulfill at the moment.
Although surviving a hit on your own might prove to be too good, it's better than let Manmelter rot forever


FortyTwoFortyTwo (commented):
Manmelter can have a change, but this change indeed doesn't really seem to help much. Especially that health on hit is only 15 without overheal, I can't see how it can help you tank another hit

PR #228, opened 2021-04-27
Third Degree

StarInAPond:
More uber gained per hit.
Meleeing Hale isn't easy, but even then, only 5% more should make a big difference.
Good concept, just needs stronger effect


FortyTwoFortyTwo (approved):

FortyTwoFortyTwo (changes requested):
Almost missed that desp need to be updated for +25% instead of +20%

PR #232, opened 2021-04-27
Sandvich and Banana

StarInAPond:
An attempt to make other lunchboxes relevant, 3 times faster taunt speed makes it practically instant

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1439,7 +1439,6 @@
 		{
 			"desp"			"Sandvich: {positive}3 times faster lunchbox consumption"
 			"attrib"		"876 ; 4.0 ; 201 ; 3.0"
-			
 		}
 		
 		"863"	//Robo-Sandvich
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1435,6 +1435,29 @@
 			}
 		}
 		
+		"42"	//Sandvich
+		{
+			"desp"			"Sandvich: {positive}3 times faster lunchbox consumption"
+			"attrib"		"876 ; 4.0 ; 201 ; 3.0"
+
```
FortyTwoFortyTwo, line 1442: Extra space

FortyTwoFortyTwo (commented):
I'm not used to faster taunt speed, but might be okay with it

FortyTwoFortyTwo (commented):
Missed another one, `3 times faster` wording isn't really similar to other weapon wording, `+200% faster` would be more fitting

FortyTwoFortyTwo (approved):

PR #235, opened 2021-04-27
Eviction Notice

StarInAPond:
An attempt to return old Eviction Notice
The reason I suggested to nerf jump height to 50% back then was too much survivability it granted, Eviction Notice has always meant to be a tool for heavies to reach a good spot, not a mean to get away, that's GRU's job.
Shouldn't be a problem now, heavy holding it will just get oneshotted

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1478,8 +1478,8 @@
 		
 		"426"	//Eviction Notice
 		{
-			"desp"			"Eviction Notice: {neutral}Active movement speed bonus is replaced with 50% greater jump height"
-			"attrib"		"524 ; 1.5 ; 851 ; 1.0"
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo (commented):
I don't remember if +70% was overpowered because of avoiding boss hits or able to reach high spots, if the latter I feel that damage vulnerability only on active wouldn't do enough and need perma effect, like max health reduction.

But in the end is this even needed, is 50% already enough?

StarInAPond:
I was the one to propose return to 50%, because heavies kept jumping over hale's heal, and when you finally hit one, it knocked him far away, still alive

FortyTwoFortyTwo:
Wonder if penalty while active still allow heavy to quickly switch back and fourth to avoid taking 1-hit, though would be more harder to do so

StarInAPond:
Could add a slower switch speed later if it's going to be a problem

FortyTwoFortyTwo (approved):

PR #236, opened 2021-04-27
Crusader's Crossbow

StarInAPond:
Legacy's Crusader's Crossbow.
If Syringe guns are used for gaining uber, this shall be medic's go-to damage weapon.
Crossbow is fairly hard to hit in the crowd, and it doesn't knockback as much as legacy one did, should work fine

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1608,8 +1608,8 @@
 		
 		"305"	//Crusader's Crossbow
 		{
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo (commented):
> If Syringe guns are used for gaining uber, this shall be medic's go-to damage weapon.

I would remove uber-on-hit on crossbow if that's your goal, otherwise this seems overboard

StarInAPond:
Agreed, however uber gain is pretty important too, how about 60% damage bonus?

StarInAPond:
changed suggestion to just 60%

FortyTwoFortyTwo (approved):

PR #239, opened 2021-04-27
The Eureka Effect

StarInAPond:
I don't think this gimmick is enough to compensate for other terrible stats, but that could be the start


FortyTwoFortyTwo (commented):
Not a fan of Eureka Effect's teleport mechanics that allows engineer to retreat from spots, already have teleporter anyway.
Weapon should just have different new stats.

PR #241, opened 2021-04-29
Jump cooldown time

StarInAPond:
This is something I wanted to discuss for a long time.
In Rewrite, Super Jump cooldown linked to the time spent charging it, with maximum cooldown being 7 seconds, and minimum 2.5. Simply put, stronger jump — longer cooldown. If I had to guess, this was done in assumption the smaller jump is = the lower value it has. On practice, this isn't the case, small jumps are just as valuable as fully changed ones, they even give same velocity.
It's easier to hit someone in the air with weak jump, they're more precise, give same speed and have lower cooldown, and when someone is really far away, you just hold button longer.

Since there's practically no difference (with weak jump being stronger), I suggest to equate all cooldowns to 7 seconds


StarInAPond:
closing because I can't edit the changes

PR #246, opened 2021-05-07
Horizontal speed multiplier on climb

StarInAPond:
Adds info about multiplier to the config and default multiplier value


FortyTwoFortyTwo:
Merged this PR to #247

PR #247, opened 2021-05-07
The Shahanshah

StarInAPond:
Gives horizontal mobility in exchange of height and a bit more self damage per climb

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1947,6 +1947,23 @@
 			}
 		}
 
+		"401"	//The Shahanshah
+		{
+			"desp"			"The Shahanshah: {positive}Greater horizontal speed on climb, {negative}less climb height, 66% extra health cost on climb"
+			
+			"attack"
+			{
+				"Tags_Climb"
+				{
+					"override"		"sniper-melee-climb"
+					
+					"height"			"600"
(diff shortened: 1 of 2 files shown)


addons/sourcemod/scripting/vsh/tags.sp
```diff
@@ -482,6 +482,7 @@ public void Tags_Climb(int iClient, int iTarget, TagsParams tParams)
 	float flHeight = tParams.GetFloat("height");
 	int iMax = tParams.GetInt("max");
 	float flDamage = tParams.GetFloat("selfdamage");
+	float flHorizontalSpeedMult = tParams.GetFloat("speed");
```
FortyTwoFortyTwo, line 485: "horizontal" may be a better name than "speed", otherwise it may be confused with "height".

Also default value 1.0 should be set here instead of requiring config to set
`tParams.GetFloat("horizontal", 1.0);`

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -415,7 +416,8 @@
 						
 						"height"		"750.0"
 						"max"			"1"
-						"selfdamage"	"15.0"	//Self-damage
+						"selfdamage"		"15.0"	//Self-damage
+						"speed"			"1.0"	//Horizontal speed multiplier
```
FortyTwoFortyTwo, line 420: No longer needed

FortyTwoFortyTwo:
We messed up this PR badly its in #259 now

PR #248, opened 2021-05-07
Bushwacka

StarInAPond:
Bushwacka is in a pretty sad state, higher climb and less self damage should fix that

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1931,7 +1931,7 @@
 		
 		"232"	//Bushwacka
 		{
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1931,7 +1931,7 @@
 		
 		"232"	//Bushwacka
 		{
-			"desp"			"Bushwacka: {positive}Higher climb height, removed damage vulnerability, {negative}233% extra health cost on climb"
+			"desp"			"Bushwacka: {positive}Higher climb height, removed damage vulnerability, {negative}166% extra health cost on climb"
```
FortyTwoFortyTwo, line 1934: Because of rounding to nearest, should be 167% instead

FortyTwoFortyTwo (approved):

PR #253, opened 2021-05-08
Big Earner

StarInAPond:
Longer speed boost to make Big Earner stand out more, backstabs are harder to perform and should be rewarded better


FortyTwoFortyTwo (commented):
I'm okay with this change, but there was talks how -25 max health penalty really doesn't change much making it a near upside weapon. Whenever if it worth to have a different downside or not

FortyTwoFortyTwo (approved):

PR #257, opened 2021-05-08
Pomson 6000

StarInAPond:
Similar to regular Pomson, it would be fitting for Pomson in Rewrite to have rage drain, unsure about the number

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1542,11 +1542,11 @@
 				{
 					"victimuber"	"0"
 				}
-					"Tags_RemoveRage"
-					{
-						"target"		"victim"
-						"amount"		"200"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1534,8 +1534,20 @@
...
+			{
+				"filter"
+				{
+					"victimuber"	"0"
+				}
+					"Tags_RemoveRage"
+					{
+						"target"		"victim"
+						"amount"		"200"
+					}
```
FortyTwoFortyTwo, line 1549: Weird tabs and spacing

FortyTwoFortyTwo (commented):
200 seems to be overboard when one beam can damage boss more than once

FortyTwoFortyTwo (approved):

PR #264, opened 2021-06-06
Disabled various pieces of dome logic when not in VSH

JillCrungus:
This should fix some issues that were stemming from dome logic still running outside of VSH, such as server console spam about the medic_alert sound not being precached (and others too, probably).

Admittedly I'm not entirely sure that all of these checks are necessary, I'm sure some of these functions won't even get called, but I don't think they hurt to have.

### DIFF
--- a/addons/sourcemod/scripting/vsh/dome.sp
+++ b/addons/sourcemod/scripting/vsh/dome.sp
@@ -246,6 +246,8 @@ stock bool TraceFilter_Dome(int iEntity, int iMask, any iData)
 
 bool Dome_Start(int iCP = 0)
 {
+	if (!g_bEnabled) return false;
+	
 	if (g_flDomeStart != 0.0)	//Check if we already have dome enabled, if so return false
 		return false;
 
@@ -406,6 +408,8 @@ public void Dome_Frame_Prepare()
 
 public void Dome_Frame_Shrink()
 {
+	if (!g_bEnabled) return;
+	
 	if (g_flDomeStart == 0.0)
 		return;
 
@@ -468,6 +472,8 @@ public void Dome_Frame_Shrink()
 
 public Action Dome_TimerBleed(Handle hTimer)
 {
+	if (!g_bEnabled) return Plugin_Continue;
+	
 	if (g_hDomeTimerBleed != hTimer)
 		return Plugin_Stop;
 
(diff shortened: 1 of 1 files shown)


addons/sourcemod/scripting/vsh/dome.sp
```diff
@@ -246,6 +246,8 @@ stock bool TraceFilter_Dome(int iEntity, int iMask, any iData)
 
 bool Dome_Start(int iCP = 0)
 {
+	if (!g_bEnabled) return false;
+
```
FortyTwoFortyTwo, line 250: `Dome_Start` gets called from `Dome_OnCapEnabled`, I'd rather have `g_bEnabled` check there instead

addons/sourcemod/scripting/vsh/dome.sp
```diff
@@ -406,6 +408,8 @@ public void Dome_Frame_Prepare()
 
 public void Dome_Frame_Shrink()
 {
+	if (!g_bEnabled) return;
+
```
FortyTwoFortyTwo, line 412: This shouldn't be needed when `Dome_Start` doesn't get called

addons/sourcemod/scripting/vsh/dome.sp
```diff
@@ -468,6 +472,8 @@ public void Dome_Frame_Shrink()
 
 public Action Dome_TimerBleed(Handle hTimer)
 {
+	if (!g_bEnabled) return Plugin_Continue;
+
```
FortyTwoFortyTwo, line 476: This shouldn't be needed when `Dome_Start` doesn't get called

FortyTwoFortyTwo (commented):
I noticed myself that `Dome_BlockOutput` should also have `g_bEnabled` check, otherwise would block some map outputs on capture.

Otherwise think that's all for enabled check for dome stuffs.

JillCrungus:
> I noticed myself that `Dome_BlockOutput` should also have `g_bEnabled` check, otherwise would block some map outputs on capture.
> 
> Otherwise think that's all for enabled check for dome stuffs.

Sorry, I'll admit that this PR is a little haphazard. I've made some changes, I think that should be it?

FortyTwoFortyTwo (approved):

PR #268, opened 2021-06-26
Better code for healing reduction

Alex-Turtle:


addons/sourcemod/scripting/vsh/bosses/boss_pyrocar.sp
```diff
@@ -320,11 +306,15 @@ methodmap CPyroCar < SaxtonHaleBase
...
 			if (!g_hPyrocarHealTimer[victim])
 			{
-				TF2Attrib_SetByDefIndex(victim, ATTRIB_LESSHEALING, 0.5);
-				TF2Attrib_ClearCache(victim);
+				TF2Attrib_SetByDefIndex(GetPlayerWeaponSlot(victim, WeaponSlot_Primary), ATTRIB_LESSHEALING, PYROCAR_HEALINGREDUCTION);
+				TF2Attrib_ClearCache(GetPlayerWeaponSlot(victim, WeaponSlot_Primary));
+				TF2Attrib_SetByDefIndex(GetPlayerWeaponSlot(victim, WeaponSlot_Secondary), ATTRIB_LESSHEALING, PYROCAR_HEALINGREDUCTION);
+				TF2Attrib_ClearCache(GetPlayerWeaponSlot(victim, WeaponSlot_Secondary));
+				TF2Attrib_SetByDefIndex(GetPlayerWeaponSlot(victim, WeaponSlot_Melee), ATTRIB_LESSHEALING, PYROCAR_HEALINGREDUCTION);
+				TF2Attrib_ClearCache(GetPlayerWeaponSlot(victim, WeaponSlot_Melee));
```
FortyTwoFortyTwo, line 314: What's the reason behind this change? I just see it giving more issues than resolving it.
Otherwise if kept, `Destroy` function also need to be updated as it still targets victim instead of weapon.

Alex-Turtle: The main reason is that the healing reduction is inconsistant and can be avoided by switching weapons.

FortyTwoFortyTwo (approved):

PR #269, opened 2021-07-06
vsh_dome_speed_duration, vsh_dome_cp_captime

StarInAPond:
It's became really popular to delay as of late, Hales delay because they know they'll have an advantage when capping at the end of the round, most red players are dead at that point.
This should solve that issue and encourage Hales to engage more instead of hiding to score cheap kills


FortyTwoFortyTwo (commented):
I'm okay into reducing `vsh_dome_speed_duration` by 30 seconds to 150 seconds, maybe reducing `vsh_dome_cp_unlock` and `vsh_dome_cp_unlockplayer` a little aswell, but nothing big.

I oppose increasing `vsh_dome_cp_captime`, especially if were going to reduce cp unlock and/or speed duration

PR #270, opened 2021-07-06
Jump cooldown time

StarInAPond:
Haven't seen discussions about previous suggestion, edited overall jump cooldown to 9 and minimum to 5.5, same reasoning as last time


FortyTwoFortyTwo (approved):
Not sure about increasing cooldown by 2 seconds, but might be okay with it.

PR #271, opened 2021-07-06
Gentle spy health

StarInAPond:
Gentle spy has pretty good rage he can use more often than most other Hales, this nerf should make him rely more on his one-shot pistol and invisibility, spy's kit is just too good to have regular health


FortyTwoFortyTwo (commented):
I really don't see Gentle Spy an issue, especially this health nerf is way behold low

PR #275, opened 2021-07-18
Another Über Ranger update

woisalreadytaken:
- Added the option for bosses to have their max health calculation consider living attackers at the moment they spawn, instead of only considering the amount of attackers that the round started with
	- Fixes the miniboss' health not scaling based on the amount of living attackers mid round
	- **Technical note:** I'm personally not happy with the `bHealthPerPlayerAliveOnly` name, but I can't think of anything better. If you have any suggestions, let me know

- The boss' and miniboss' Mediguns now have a 60% übercharge rate penalty
- Replaced the boss' melee weapon with an Übersaw, which gains 10% über on hit
	- These changes are being made to discourage the boss and his friends from hiding away from action for extended periods of time

- Updated the boss' model and textures
	- Compressed textures so they weigh 4 MB less
	- Fixed some scuffed details in the textures
	- Made the backpack get the über texture when übered
	- Fixed non-cosmetic-item textures having 2 layers of very slight darkness being stacked on each other
	- Put the cosmetics into the model, instead of handling them through code
		- Fixes modifiers not changing colors of the cosmetics
	
- Increased the boss' health to low (700 + 650 HP per player), matching Painis Cupcake


addons/sourcemod/scripting/include/saxtonhale.inc
```diff
@@ -239,6 +239,13 @@ methodmap SaxtonHaleBase
 		public native set(bool val);
 	}
 	
+	// Retrieve/Set whether the boss' max health calculation should consider currently-living players or the playercount at the beginning of the round
+	property bool bHealthPerPlayerAliveOnly
```
FortyTwoFortyTwo, line 243: I think `bHealthPerPlayerAlive` without `Only` is best to name it, there already property similarly named `iHealthPerPlayer`

FortyTwoFortyTwo (approved):
New models to be uploaded to servers before merging this

PR #316, opened 2021-11-12
Eureka Effect changes

Dencube:
The Eureka Effect has been outclassed by other engineer wrenches for a while. Hopefully 20 metal every 5 seconds and slightly reduced construction speed penalty is enough to make it a viable option in vsh. In order to not make it too strong, the metal penalty from dispensers/pickups will stay.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1767,8 +1767,8 @@
 		
 		"589"	//The Eureka Effect
 		{
-			"desp"			"The Eureka Effect: {positive}+20 metal every 5 seconds, reduced construction speed on hit penalty"
-			"attrib"		"113 ; 20.0 ; 93 ; 0.65"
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1765,6 +1765,12 @@
 			"prefab"		"140"
 		}
 		
+		"589"	//The Eureka Effect
+		{
+			"desp"			"The Eureka Effect: {positive}+20 metal every 5 seconds, reduced construction speed on hit penalty"
```
FortyTwoFortyTwo, line 1770: reducing construction on hit feels a bit redundant as it would do very little difference. But if were to keep it, should add `to 35%` at the end to mention exact valve change

Dencube: removed the construction on hit change

FortyTwoFortyTwo (approved):

PR #318, opened 2021-11-12
Add Tags_IgnitePlayer + some new filters for minimum/maximum damage

infinitephantasm:
originally these additions were intended as an assistance to a change to pyro's shotgun, however after realizing the trouble with reskins & similarly-classes shotguns, i gave up. however, the changes added nonetheless might prove useful in other weapon changes.

`minimumdamage` filter checks to see if attacker dealt above a certain amount of damage

`maximumdamage` filter checks to see if attacker dealt below a certain amount of damage

`Tags_IgnitePlayer` is essentially just a proxy to `TF2_IgnitePlayer`. it sets people on fire.

if there are any tab spacing issues blame notepad++ :v

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -37,8 +37,8 @@
 //- attackweapon          - Is attacked weapon in slot X, only works on 'attackdamage' and 'takedamage'
 //- aim                   - Is client aiming at boss
 //- sentrytarget          - Is client's sentry targeting at boss
-//- damagemax		  - Did client deal less than X amount of damage
-//- damagemin		  - Did client deal more than X amount of damage
+//- damagemax             - Did client deal less than X amount of damage
+//- damagemin             - Did client deal more than X amount of damage
 //- damagetype            - Does damagetype have X, only works on 'attackdamage' and 'takedamage'
 //- damagecustom          - Is damagecustom X, only works on 'attackdamage' and 'takedamage'
 //- backstabcount         - X consecutive backstabs required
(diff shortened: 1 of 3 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -41,6 +41,8 @@
 //- damagecustom          - Is damagecustom X, only works on 'attackdamage' and 'takedamage'
 //- backstabcount         - X consecutive backstabs required
 //- feigndeath            - Does client have feign death ready
+//- minimumdamage		  - Did client deal more than X amount of damage
+//- maximumdamage		  - Did client deal less than X amount of damage
```
FortyTwoFortyTwo, line 45: Think it would be better to have it named `damagemin`/`damagemax` so it would be consistent for `damagetype` and `damagecustom` having `damage` first in name, and `min`/`max` is named in params.
Same to enums etc in `tags_filter.sp`

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -175,6 +177,8 @@
 //
 //Tags_MakeBleed          - Adds bleed to target
 //- duration                - Bleed duration
+//Tags_IgnitePlayer		  - Sets target on fire
```
FortyTwoFortyTwo, line 180: Missing a line comment `//` above to make it consistent

addons/sourcemod/configs/vsh/vsh.cfg
infinitephantasm: should be adjusted, hope i didn't miss anything in `tags_filter.sp`

addons/sourcemod/configs/vsh/vsh.cfg
infinitephantasm: resolved

FortyTwoFortyTwo (approved):

PR #323, opened 2021-11-23
Crit-a-Cola rework

Dencube:
Changes crit-a-cola into a mini version of buff banner with a 30 second recharge time

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -640,6 +640,38 @@
 			"prefab"		"46"
 		}
 		
+		"163"	//Crit-a-Cola
+		{
+			"desp"			"Crit-a-Cola: {positive}Grants 8 temporary clip size with 10x fire rate {negative}30 second recharge time"
+			"attrib" 		"801 ; 30.0"
+			
+			"lunchbox"	//Called when client consumes lunchbox
+			{
+				"Tags_AddAttrib"
+				{
+					"target"		"primary"	//Primary weapon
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -640,6 +640,38 @@
 			"prefab"		"46"
 		}
 		
+		"163"	//Crit-a-Cola
+		{
+			"desp"			"Crit-a-Cola: {positive}Grants 8 temporary clip size with 10x fire rate {negative}30 second recharge time"
```
FortyTwoFortyTwo, line 645: Missed a comma before negative comments

FortyTwoFortyTwo (commented):
Played around with different scatterguns damaging boss right in front of it.

Scattergun deals about same damage to rocket launcher with buff banner (~2.4k dmg), however using scout drinks is much easier than build and using buff banner, so I'm not sure whenever crit-a-cola need a nerf, or buff banner needs a buff.

As for other scatterguns, rest is weaker than scattergun because of less base clip size. I'm not sure what to do with this, when taking into account whenever if weapon is fully reloaded or not.

Back Scatter deals like 2-3x less damage than Scattergun, I'd just remove crit becomes minicrit attribute.

addons/sourcemod/configs/vsh/vsh.cfg
Dencube: fixed the missing comma

Dencube:
removed crit to minicrits and slightly lowered the stats

FortyTwoFortyTwo (approved):

PR #327, opened 2021-12-13
scout bonus damage

StarInAPond:
If scare from all rages gets removed, this buff will be unnecessary


FortyTwoFortyTwo:
I'm not into nerfing scout's poor damage even further in favour of more jumping around

PR #329, opened 2021-12-13
Eyelander

StarInAPond:
Experimental change, since Eyelander had it's speed removed, I want to see new health stacking gimmick to be a little bit better, let it be proper alternative playstyle

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1396,12 +1396,12 @@
 					"index"			"26"		//Max health on wearer
 					
 					"math"			"add"		//Add value everytime this is called
-					"value"			"15"		//15 max health on every hit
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -1396,12 +1396,12 @@
 					"index"			"26"		//Max health on wearer
 					
 					"math"			"add"		//Add value everytime this is called
-					"value"			"15"		//15 max health on every hit
+					"value"			"25"		//2	5 max health on every hit
```
FortyTwoFortyTwo, line 1399: weird spacing between 2 and 5

FortyTwoFortyTwo (commented):
Description needs updating for +25 instead of +15

StarInAPond:
how could that happen... 
fixed

FortyTwoFortyTwo (approved):

PR #330, opened 2021-12-13
Huntsman

StarInAPond:
Back to the old critical huntsman, shouldn't be too strong since recently bosses received health buff :)


FortyTwoFortyTwo:
I'm not into making it crit as it was very strong before, I'd be okay if explosive damage and/or radius is buffed instead

StarInAPond:
What about that? Serious damage and decent radius increase, should still be weaker on direct impact, but splash damage spam sounds fun

FortyTwoFortyTwo (approved):

PR #337, opened 2021-12-19
Config rewording

woisalreadytaken:
Lots of changes by lots of contributors done in a considerable amount of time tend to lead to inconsistency (and in some cases even slightly wrong info 😮 !) so I reworded stuff to be a bit more consistent

Anyone who wants to bother to read this stuff can let me know if I missed something or if you don't like what I typed and want to suggest something else, it's cool, we chillin

Just remember, this has no gameplay changes at all

Changes include:
- some color code rearranging (firing x instead of y should be neutral and not positive)
- changing "movement speed" to "move speed" because that's what valve apparently calls it in all cases, plus it's shorter
- the addition of a single display only "no reload necessary" attribute to the bison
- removing redundant +/- on stats that are already shown as bonuses/penalties/faster/slower/etc
- "boss's" > "boss'" to consistently match with "players'"
- "up to" removed from backup because it has been only spawning 3 zombies with a recent-ish change, but kept in bfb because its charge meter sometimes fills up with 380 damage or some shit like that for a reason i forgot (happens in stock tf2 to a lesser extent too)
- fixing some typos
- some more stuff

As far as things I haven't missed go, I'm just not happy with the jarate description because the words "rage" and "drain" show up twice and I want to keep it consistent with the pomson, if you have any suggestions let me know

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -211,7 +211,7 @@
 		{
 			"Primary"
 			{
-				"desp"		"Primary: {positive}Crit rockets when aiming directly at the boss"
+				"desp"		"Primary: {positive}Crits when aiming directly at the boss"
 				
 				"attack"
 				{
@@ -237,7 +237,7 @@
 		{
 			"Primary"
 			{
-				"desp"		"Primary: {positive}+200% afterburn damage bonus, {negative}12 seconds airblast cooldown"
+				"desp"		"Primary: {positive}200% afterburn damage bonus, {negative}12 seconds airblast cooldown"
 				"attrib"	"71 ; 3.0 ; 171 ; 0.0"
 				
 				"spawn"
@@ -430,7 +430,7 @@
 			
 			"Melee"
 			{
-				"desp"		"Melee: {neutral}Backstabs deal 1000 damage"
+				"desp"		"Melee: {neutral}Backstab deals 1000 damage"
 				
 				"attackdamage"
 				{
@@ -552,7 +552,7 @@
 		
 		"1103"	//Back Scatter
 		{
-			"desp"			"Back Scatter: {positive}Fires fast rockets instead of bullets, crits when aiming at boss, {negative}deals less damage"
+			"desp"			"Back Scatter: {positive}Crits when aiming directly at the boss, {neutral}fires fast rockets instead of bullets"
 			"attrib"		"280 ; 2.0 ; 2 ; 8.0 ; 103 ; 1.33"
 			"attack"
 			{
@@ -612,7 +612,7 @@
 		
 		"46"	//Bonk! Atomic Punch
 		{
-			"desp"			"Bonk: {neutral}Replaces bonk effect into defense buff"
+			"desp"			"Bonk: {neutral}Replaces bonk effect with a defense buff for 8 seconds"
 			
 			"think"
 			{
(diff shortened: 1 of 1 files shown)


addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -552,7 +552,7 @@
 		
 		"1103"	//Back Scatter
 		{
-			"desp"			"Back Scatter: {positive}Fires fast rockets instead of bullets, crits when aiming at boss, {negative}deals less damage"
+			"desp"			"Back Scatter: {positive}Crits when aiming directly at the boss, {neutral}fires fast rockets instead of bullets"
```
FortyTwoFortyTwo, line 555: Does make more sense to have neutral first before positive for this and anywhere else

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -908,7 +908,7 @@
 		
 		"226"	//Battalion's Backup
 		{
-			"desp"			"Battalion's Backup: {positive}Spawns up to 3 Zombie Scouts on use"
+			"desp"			"Battalion's Backup: {positive}Spawns 3 Zombie Scouts on use"
```
FortyTwoFortyTwo, line 911: "up to" is still true when there not enough dead players or revival preference to respawn, but I'm not sure if that should be enough to keep it

addons/sourcemod/configs/vsh/vsh.cfg
```diff
@@ -2037,7 +2037,7 @@
 		
 		"58"	//Jarate
 		{
-			"desp"			"Jarate: {neutral}Replaces jarate effect into a rage drain, removing equivalent of 400 damage"
+			"desp"			"Jarate: {neutral}Replaces jarate effect with a rage drain, drains 400 damage worth of rage"
```
FortyTwoFortyTwo, line 2040: `Replaces jarate effect with 400 damage worth of rage drain`? Not sure what else it can be worded

FortyTwoFortyTwo (commented):
> removing redundant +/- on stats that are already shown as bonuses/penalties/faster/slower/etc

TF2 attributes have +/- even if wording is here, though not all have it so eh. Not sure if all should have it or match with TF2, but still would like to keep some of it.

addons/sourcemod/configs/vsh/vsh.cfg
woisalreadytaken: that's true, i forgot about that

woisalreadytaken:
> TF2 attributes have +/- even if wording is here, though not all have it so eh. Not sure if all should have it or match with TF2, but still would like to keep some of it.

nah not all of it is removed, if something is like `50% damage bonus` then there won't be a + sign because the word "bonus" is already there, but there's still stuff like `+50% health from packs` because it wouldn't be clear without the sign

though if you'd like to make it really simple, could just remove "bonus/penalty/etc" and it'd still be correct (`+50% damage`, `-20% move speed`, and so on)

FortyTwoFortyTwo:
> or just adding signs to everything that makes sense is fine too, as long as it's not too inconsistent it's cool

Go for signs, it's just one extra symbol that can help read which way attribute goes.

woisalreadytaken:
ok 42

addons/sourcemod/configs/vsh/vsh.cfg
woisalreadytaken: I'll just leave this one as is, it's not too bad

FortyTwoFortyTwo (approved):

PR #358, opened 2022-02-06
a different way to balance the fan

jewishbukkake:
all i want is for the fan to be viable if this is still not good i will keep on trying other ways


FortyTwoFortyTwo:
I really left this pending for so long. I still don't want to take the risk on weapon potentially becoming a problem in the future.

Also, the desc was formatted incorrectly to have buff labelled as negative.

PR #361, opened 2022-03-06
Degreaser: undo current attributes

woisalreadytaken:
This is a major change.

being completely honest with you, I don't think the Degreaser we have right now is healthy for the gamemode. being sent across a room (or half the map if the pyro is standing on a high edge), per pyro, is not a fun thing for bosses to deal with.

that said, the Degreaser with no changes is pretty shit in this gamemode so this is some sort of discussion point: how to make it better?

if anyone has a suggestion, let us know somewhere, but this PR has a suggestion of my own: give it the passive weapon switch speed that it had until Tough Break. it's pretty tame and doesn't sound like much, but people could hopefully have an easier time comboing it with something, as it's the point of the Degreaser.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1033,17 +1033,8 @@
 		
 		"215"	//Degreaser
 		{
-			"desp"			"Degreaser: {positive}+100% airblast push force, {negative}+100% airblast cooldown"
-			"attrib"		"255 ; 2.0"
-			
-			"spawn"
-			{
-				"Tags_Airblast"
-				{
-					"override"		"pyro-primary-airblast"
-					"cooldown"		"24"
-				}
-			}
+			"desp"			"Degreaser: {positive}Passive +65% faster weapon switch speed"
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo:
I'm not sure about the idea for faster weapon switch, could that make it more annoying to bosses when combining with secondary weapon? I agree that Degreaser needs to be changed, but I don't have any other suggestions.

woisalreadytaken:
> could that make it more annoying to bosses when combining with secondary weapon?

i don't think it can be that bad

if you mean the scorch shot, at least it'd be most likely be used for self defense and wouldn't be too different from using it without the degreaser unless you want to sprinkle in some minor flamethrower damage

if you mean the manmelter, if it can do enough damage to be concerning between weapon switches in the first place, that's more like a problem with that weapon instead of the degreaser

the only thing that could not play well that i can think of is axtinguisher damage output if the pyro has a medic with him, but that needs some brain power to make good use of, at least

FortyTwoFortyTwo:
Went ahead with just simply removing changes without any new stats. It could have a buff if there a better idea, but best to have a weak weapon than a strong weapon.

PR #363, opened 2022-03-10
Gunslinger buff

Dencube:
Overshined by other engineer wrenches. These changes should make it more viable while not being too strong. (hopefully)


Dencube:
changed my mind. this wasn't a good idea.

PR #367, opened 2022-05-03
Demo Boots Buff

Dencube:


FortyTwoFortyTwo (commented):
I'm not sure myself if blast damage would be any helpful to it, but I would like to see speed bonus changes be removed if were to go for it, to not clutter too much changes in one.

FortyTwoFortyTwo:
Focused into not adding too many different changes now.

PR #368, opened 2022-05-03
Huo Long Heater buff

Dencube:


FortyTwoFortyTwo (commented):
I honestly see this change as a nerf, considering that this won't have extra afterburn damage like flamethrower, and afterburn don't stack with other afterburn sources.

~~Also what is attribute `795`/`damage bonus vs burning` here for? seems pointless.~~ Nevermind there is a reason.

FortyTwoFortyTwo:
Still prefer current changes.

PR #377, opened 2022-10-21
Gameplay & QoL changes to HHH Jr.

woisalreadytaken:
Gameplay changes:
- HHH Jr.
	- Fixed rage not affecting bosses (including Zombie Scouts), in which cases it'll hurt them, but not regen your own hp
	- Reduced super rage radius from 150% (600 units) to 125% (500 units) of regular rage radius
	- Removed occasional view disorientation from the rage's random effect pool
	
It's all up to discussion!

QoL changes:
- Teleport-Swap ability
	- Added a teleporting sound on both ends of teleport-swap
	- Fixed the charge meter not going down on button release when failing to trigger due to an external source (like while stunned or airborne)
		- Such external sources will now be displayed in a little hint message to the user if the ability is blocked
	- The on-button-hold text has been changed to "Teleport-swap: [round number]％. Look up and release right click to teleport."
		- Note that a full-width percentage sign was used here. It looks a little bit cut off (see image at the bottom), but it's better than having to dance around the hyper scuffedness of text with a bunch of formatting
	- The cooldown text has been changed to "Teleport-swap is on cooldown for x second(s)!"

For the meter/text changes, every similar ability should get the same treatment. I'm just not sure if that should be done in this or another pull request, as this one is entirely related to HHH Jr.

image

### DIFF
--- a/addons/sourcemod/scripting/vsh/abilities/ability_rage_ghost.sp
+++ b/addons/sourcemod/scripting/vsh/abilities/ability_rage_ghost.sp
@@ -75,7 +75,7 @@ public void RageGhost_OnRage(SaxtonHaleBase boss)
 	
 	//Thirdperson
 	SetVariantInt(1);
-	AcceptEntityInput(boss.iClient, "SetForcedTauntCam");
+	AcceptEntityInput(iClient, "SetForcedTauntCam");
 }
 
 public void RageGhost_OnThink(SaxtonHaleBase boss)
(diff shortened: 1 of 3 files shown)


addons/sourcemod/scripting/vsh/abilities/ability_rage_ghost.sp
```diff
@@ -279,44 +284,29 @@ public void RageGhost_OnThink(SaxtonHaleBase boss)
 			//Other random effects
 			for (int i = 0; i < iLength; i++)
 			{
-				bool bEffectDone = false;
-				
 				//Attempt use random slot
 				if (GetRandomInt(0, 1))
```
FortyTwoFortyTwo, line 288: Do we want to remove this chance? Since this is here to roll for either weapon switch or random angles. Could increase cooldown in trade with it.

woisalreadytaken: Yeah I misunderstood this, by glancing over I thought it was just a random chance to trigger either effect at all. Removed it and increased the effect cooldown to 2.5 seconds so it happens 4 total times (as it does that at the beginning of the rage as well).

FortyTwoFortyTwo (approved):

PR #379, opened 2022-11-27
Add music volume compatibility to boss themes

woisalreadytaken:
This PR allows players to adjust boss theme volume with TF2's music volume slider. Only applies to boss themes, so rage music and some song-voicelines like Demopan's win theme are unaffected.

Custom and future bosses will need to manually make use of the # prefix character and `PrepareMusic` when precaching sounds to be affected by this change.

### DIFF
--- a/addons/sourcemod/scripting/vsh/bosses/boss_yeti.sp
+++ b/addons/sourcemod/scripting/vsh/bosses/boss_yeti.sp
@@ -173,7 +173,7 @@ public void Yeti_Precache(SaxtonHaleBase boss)
 {
 	PrecacheModel(YETI_MODEL);
 	
-	PrecacheSound(YETI_THEME);
+	PrepareMusic(YETI_THEME, false);
 	
 	for (int i = 0; i < sizeof(g_strYetiRoundStart); i++)PrecacheSound(g_strYetiRoundStart[i]);
 	for (int i = 0; i < sizeof(g_strYetiWin); i++)PrecacheSound(g_strYetiWin[i]);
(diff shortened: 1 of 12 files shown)


FortyTwoFortyTwo:
Can we just make it so that `g_hTimerBossMusic` automatically get `#` prefix when calling `GetMusicInfo` without needing to add one to each bosses with `PrepareMusic`?

woisalreadytaken:
done, though with this change, custom bosses NEED to use `PrepareMusic` to get their theme to work at all

addons/sourcemod/scripting/vsh/bosses/boss_yeti.sp
```diff
@@ -1,5 +1,5 @@
 #define YETI_MODEL "models/player/kirillian/boss/yeti_modded.mdl"
-#define YETI_THEME "ui/gamestartup29.mp3"
+#define YETI_THEME "#ui/gamestartup29.mp3"
```
FortyTwoFortyTwo, line 2: Missed this to revert the #

addons/sourcemod/scripting/vsh/stocks.sp
```diff
@@ -966,6 +966,25 @@ stock void PrepareSound(const char[] sSoundPath)
...
+	char s[PLATFORM_MAX_PATH];
+	FormatEx(s, sizeof(s), "#%s", sSoundPath);
+	PrecacheSound(s, true);
+	
+	if (!bCustom)
+		return;
+	
+	if (ReplaceString(s, sizeof(s), "#", "sound/") != 1)
+	{
+		PrintToServer("[VSH REWRITE] PrepareMusic could not prepare %s: filepath must not have any '#' characters.", sSoundPath);
```
FortyTwoFortyTwo, line 981: Use `LogError` to make it more easily noticeable that this is not supposed to happen, vsh prefix also not needed.

FortyTwoFortyTwo (commented):
I was not aware that `PrecacheSound` also need `#` to make it work(?), but still prefer to go this way. Could add `SaxtonHale_PrepareMusic` to `saxtonhale.inc` for sub-plugins to be able to easily use it, but not needed.

PR #384, opened 2023-05-30
Clear unloaded plugin properly

Jugheadbomb:
As SM 1.11 is stable, let's use **OnNotifyPluginUnloaded**.


FortyTwoFortyTwo (approved):
Looks good with quick tests I've done, thanks!

PR #388, opened 2023-07-27
Horsemann: Add eye glow particles

Jugheadbomb:
image
P.S.: Not image from VSH directly, just to show how eyeglow looks


addons/sourcemod/scripting/vsh/bosses/boss_horsemann.sp
```diff
@@ -120,6 +120,8 @@ public void Horsemann_OnSpawn(SaxtonHaleBase boss)
 	264: melee range multiplier (tf_weapon_sword have 37% extra range)
 	551: special taunt
 	*/
+
+	Horsemann_CreateEyeGlow(boss);
```
FortyTwoFortyTwo, line 124: Can the eye glow entity stack if boss were to keep suiciding during setup time? Don't know if `Horsemann_Create` would work there, or would have to kill existing glow ents.

PR #389, opened 2023-08-17
YER: Deal 500 damage and 6500 on final stab

Mikusch:
Solves the age-old problem of YER spies being completely useless to the team without affecting their damage too much


FortyTwoFortyTwo (changes requested):
There desp to update on new values, but I also would've lowered final stab to 6000, nicer number and otherwise a direct buff

FortyTwoFortyTwo (approved):

PR #390, opened 2023-10-08
Add CTakeDamageInfo struct for damage hooks

Jugheadbomb:
This PR is used to replace `SDKHook_OnTakeDamageAlive` with detour `CTFPlayer::OnTakeDamage_Alive` for accessing CTakeDamageInfo object, which is used to structure damage hooks code a bit, and also to access `m_eCritType`, for example.
P.S.: PR also adds `crittype` filter and param for `(attack/take)damage` tags

Previously, I made these changes for my own fork of vsh - all worked fine. Can't say guaranteed that this PR will make it work fine, need more tests.

This PR also updates dhooks code to use methodmaps.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -28,6 +28,7 @@
 //- multiply              - From 'attackdamage' and 'takedamage', multiplies damage by X
 //- max                   - From 'attackdamage' and 'takedamage', max possible damage to be made
 //- damagetype            - From 'attackdamage' and 'takedamage', damage types to add/remove
+//- crittype              - From 'attackdamage' and 'takedamage', set damage crit type
 //- attackcrit            - From 'attack' key, set whenever if weapon should crit or not
 //- healbuilding          - From 'heal' key, enable/disable healing building
 //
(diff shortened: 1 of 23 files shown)


FortyTwoFortyTwo:
While I appreciate the effort into it, I'm not sure if it's even worth to have offsets and gamedatas just for some extra variables that may or may not be useful, as this also needs updating on any sub-plugins that added its own boss. Do you have examples on how this could be useful for?

Jugheadbomb:
Seems you are right, especially on part about requiring updating all sub-plugins. I don't have any useful examples on this, so you can close PR if you don't consider it necessary. Just thought it would do some more convenient code using struct to maintain this code with. Thanks for feedback.

Kenzzer:
Should this PR somehow go through. I'd merely suggest to pick [SourceTools](https://github.com/TF2-DMB/CBaseNPC) since well it has CTakeDamageInfo mm, so the gamedata maintenance wouldn't fall on you guys. Redsun already has it in its dependencies, and most community servers do as well

But its quite the chunky dependency to pull in, however if VSH-Rewrite keep piling new features request this could be something worth considering.

FortyTwoFortyTwo:
It can certainly be used in the future if there were to be more reason to use it, just that right now there isn't really a need to do it.
If this were to ever come, I'd also like to see methodmap get used instead of enum struct, as struct between plugins can break whenever there an update to it.

PR #395, opened 2023-12-24
Remove unnecessary downloads

Dispenz0r:
Safely removes sw.vtx files as they're not used by the game, should reduce overall download times.


FortyTwoFortyTwo (changes requested):
Surprised there isn't a stock made to auto-add all of the model files...
Anyway, there `models/kirillian/brsphere_huge.sw.vtx` you forgot to remove from `AddFileToDownloadsTable` aswell

FortyTwoFortyTwo (approved):

PR #396, opened 2024-01-03
Back Scatter: crits from behind and at close range, and OnTakeDamage support for bosses/tags

woisalreadytaken:
For the public: the Back Scatter will crit bosses from behind and if they are within 512 hammer units. Cool!

The more technical information:
* Renamed the OnTakeDamage and OnAttackDamage boss functions to OnTakeDamageAlive and OnAttackDamageAlive respectively, as they were already using the OnTakeDamageAlive hook.
* The OnTakeDamage and OnAttackDamage boss functions still exist, but now happen in the OnTakeDamage hook.
  * Allows boss code to change damagetype so hits they deal/take can be proper crits. As well as change the damage inflictor if you so wish.
  * **This might break custom bosses that currently opt into them!!!** They (probably) won't crash the server or anything, they'll just be using the wrong damage hook.
* Damagetypes from damage tags are now handled in OnTakeDamage, so you can deal proper crit damage through the config.
* Added the "hitfrombehind" tag filter, only goes through if the victim is hit from behind (duh), mimicking the closerange_backattack_minicrits attribute's code, and within the specified distance in hammer units.

I just hope I didn't miss anything. Did some testing myself and it all worked fine, but you never know.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -43,6 +43,7 @@
 //- damagetype            - Does damagetype have X, only works on 'attackdamage' and 'takedamage'
 //- damagecustom          - Is damagecustom X, only works on 'attackdamage' and 'takedamage'
 //- backstabcount         - X consecutive backstabs required
+//- hitfrombehind         - Victim is attacked from behind and at close range, value is maximum distance between targets (ignored if at or below 0), only works on 'attackdamage' and 'takedamage'
 //- feigndeath            - Does client have feign death ready
 //- victimuber            - Is victim ubered or not, only works on 'attackdamage' and 'takedamage'
 //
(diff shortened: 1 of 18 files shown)


FortyTwoFortyTwo (commented):
I believe telefrag code needs to be moved to `OnTakeDamage` so it can be handled before tags for Eureka.

Also does any boss functions still need to use alive hook, or can all or most of it be moved to non-alive hook?

woisalreadytaken:
> Also does any boss functions still need to use alive hook, or can all or most of it be moved to non-alive hook?

I don't think any of the bosses/modifiers/abilities need to use the alive hook, but Pyrocar would change a very little bit if moved from it.

FortyTwoFortyTwo:
> I don't think any of the bosses/modifiers/abilities need to use the alive hook, but Pyrocar would change a very little bit if moved from it.

Would use `OnTakeDamage` to all of it, while Pyrocar can keep the alive if it's necessary

FortyTwoFortyTwo (approved):

PR #406, opened 2024-06-02
Replace Saxton Stun Rage with Berserk Rage and Lunge Ability

Batfoxkid:
- Adds rage ability RageAttributes
  - Applies temporary attribute multiplier to a weapon
- Adds passive ability Lunge
  - Leaps forwards with a damage AOE
  - Optionally consumes rage to use

- Changes Saxton Hale
  - Added +70% firing speed during rage
  - Added -50% (-25% in Super Rage) damage during rage
  - Added x1.3 movement speed during rage
  - Added knockback immunity during rage
  - Increased Super Rage duration from 7.5s -> 8.0s
  - Added a Lunge passive
  - Updated model to add a Lunge animation
  - Removed stun on rage

https://github.com/redsunservers/VSH-Rewrite/assets/15969488/1ef80211-f44b-478f-a1be-409b3449e138

https://github.com/redsunservers/VSH-Rewrite/assets/15969488/eaebeeb3-64f7-484c-a71d-f1f4e014eefc


addons/sourcemod/scripting/vsh/bosses/boss_hale.sp
```diff
@@ -159,6 +178,45 @@ public void SaxtonHale_OnSpawn(SaxtonHaleBase boss)
 	*/
 }
 
+public void SaxtonHale_OnRage(SaxtonHaleBase boss)
+{
+	boss.flSpeed *= 1.3;
+	g_bHaleSpeedRage[boss.iClient] = true;
+}
```
FortyTwoFortyTwo, line 185: This might be able to be stacked and break if rage is done before current rage isn't finished

Batfoxkid:
> For the model, is this just an update we're currently using from the same creator, and we're allowed to use it?

Yes it is, same model with an added animation replacement on kart_dash

FortyTwoFortyTwo (approved):

PR #412, opened 2024-11-19
Fix Modifier Particle Rendering

Batfoxkid:
* Fixes the incorrect SetTransmit function
* Made the particle always transmit to players on spawn for 0.2 seconds (to make it properly render)
* Cloak and disguise conditions remove/readd particle effects


addons/sourcemod/scripting/saxtonhale.sp
```diff
@@ -1159,11 +1159,18 @@ public void TF2_OnConditionAdded(int iClient, TFCond nCond)
 	if (SaxtonHale_IsValidBoss(iClient))
 	{
 		SaxtonHaleBase(iClient).CallFunction("OnConditionAdded", nCond);
```
FortyTwoFortyTwo, line 1161: Not sure why milked code is here when it could've been in base_boss and making use of this call along with particles removed, same to conds removed.

I'll still merge it regardless if this is changed or not.

PR #416, opened 2025-01-29
Restrict "scoutknight" to Three-Rune Blane only

GeeNoVoid:
Keep the Boston Basher as it normally is, but let people equip the Three-Rune Blade if they _really_ want to go melee only as scout.


FortyTwoFortyTwo:
Still would rather have Boston Basher and Three-Rune Blane keep the same stats together.

Don't care much if both weapons have the scoutknight gimmik removed.

FortyTwoFortyTwo (approved):

PR #420, opened 2025-02-18
Enable Jumper Weapons

Waifers:
surely nothing bad will come from it :clueless:

also theyre given marked for death to start with so players caught holding them will most likely die


FortyTwoFortyTwo (commented):
There seems to be a consensus that nerfing/reverting extra max ammo would be a better option to nerf jumper weapons

PR #422, opened 2025-04-08
Teuton Knights

Batfoxkid:
Adds a new setting `vsh_teuton_enable` which makes dead players spawn in as weak minion. This can be used with or as an alternative to the Dome. Dead players can't interact with triggers, collide with players, be healed, and can only deal 30 damage to enemies at a time. Dead players are slightly slower the base boss speed and can wall climb to have the ability to catch up to a boss but not constantly able to deal damage. This helps engagement for dead players and prevents the boss from camping a certain spot.

Yes this may undergo balance as this addition buffs up red-side, with a possible solution of a slight increase to boss base health. But the main goal of this is a better alternative as a dome to allow the map to be used more rather then centered on the control point and provides engagement to dead players.

- Adds "Teuton" minion boss
- Increased default dome time
- Minions can now be summoned

This is feature complete, just needs to see if there's any specific problems they may show up, and of course live balance testing.

### DIFF
--- a/addons/sourcemod/scripting/saxtonhale.sp
+++ b/addons/sourcemod/scripting/saxtonhale.sp
@@ -955,6 +955,7 @@ public void OnMapStart()
 		g_iSpritesGlow = PrecacheModel("materials/sprites/glow01.vmt", true);
 		
 		Dome_MapStart();
+		Teuton_MapStart();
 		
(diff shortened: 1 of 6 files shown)


FortyTwoFortyTwo:
I don't see myself the need to take this approach, dome concept is fine to continue with it.

PR #423, opened 2025-04-20
Enforcer Rework

Waifers:
Enforcer Rework
Has two stats from Uber update: 20% damage bonus at all times (replacing conditional damage bonus), and cloak delay Has bleed
No longer pierces resists or gain crits when disguised Still has critical damage fall-off and firing rate penalty

Specializes the Enforcer more into a higher-damage revolver at closer ranges with increased risk of getting caught. Could probably be readjusted but the current core functionality of Enforcer relying on disguising doesn't really belong in VSH imo and resist piercing could end up having some niche, scuffed interactions with bosses in the future.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -2145,23 +2145,8 @@
 		
 		"460"	//Enforcer
 		{
-			"desp"			"Enforcer: {positive}Receive crits while disguised, {negative}critical damage is affected by range, no minicrits"
-			"attrib"		"868 ; 1.0"
-			"minicrit"		"0"
-			
-			"think"
-			{
-				"filter"
-				{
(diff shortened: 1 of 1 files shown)


Waifers:
Right now, theres pretty much two main ideas for the direction the Enforcer stats could be taken:

1. Just full revert to Uber Update stats (20% increased damage at all time, but increased cloak delay).

2. Go into a more gimmicky direction and focus on higher powered, but limited shots (only one bullet in a clip/longer reload/less ammo in general). Could probably keep the critical damage falloff to make sure it cant be used to snipe. 
Maybe at least ~x7 damage per shot compared to base revolver and also having the shots apply bleed 🥴

FortyTwoFortyTwo:
Don't care much on which way this weapon goes, as long as the the wording on changelist don't go overboard compared to other weapons.

PR #427, opened 2025-05-09
Shotgun accuracy buff

StarInAPond:
Makes all shotguns besides Frontier Justice way more accurate in an attempt to make them viable. Was unsure whether to keep mini-crits on heavy, but I want to see if it will be viable alternative to minigun, something heavy uses even less than other classes


woisalreadytaken:

PR #428, opened 2025-05-10
Add blast jumping condition to some boss movement abilities + add a new config filter

woisalreadytaken:
This PR adds the blast jumping condition to bosses when they use the following abilities:
- Brave Jump
- Saxton Hale's Lunge
- Horseless Headless Horsemann Jr.'s Wall Climb
- Bonk Boy's Dash
...but purposefully not to:
- Jumper Modifier's Leap (because it's just jumping)
- Merasmus' Super Jump Spell (because it's a stock TF2 mechanic that's not as easy to modify)

...until they land, which in practice... doesn't actually do anything. But! The config is getting a new attackdamage filter, `victimcond`, which allows letting the Direct Hit and Reserve Shooter minicrit bosses using these abilities, as if they were blast jumping.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -34,6 +34,7 @@
 //
 //List of possible filters
 //- cond                  - Does client have cond id X
+//- victimcond            - Does the victim have cond id X, only meant for 'attackdamage', use 'cond' for 'takedamage'
 //- activeweapon          - Is client active weapon X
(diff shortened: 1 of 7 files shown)


FortyTwoFortyTwo (commented):
Would it be better to give attacker a minicrit buff instead of to victim a mark of death so only attacker would have the effect?

woisalreadytaken:
Yeah that works

FortyTwoFortyTwo (approved):

PR #431, opened 2025-05-10
Pyroknight VSH (Sharpened Volcano Fragment adjustment)

Waifers:
Changes the Sharpened Volcano Fragment because it has the problem of competing with the Flamethrower.
Makes Pyro tankier with a stronger melee that could be boosted further at the cost of not being able to use primaries or secondaries.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1107,10 +1107,27 @@
 		
 		"348"	//Sharpened Volcano Fragment
 		{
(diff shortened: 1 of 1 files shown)


Mikusch:
The hell? Why?

Mikusch:
Overriding this, we don't need more weapons that straight up lock you out of your entire kit.

FortyTwoFortyTwo:
You have no authority to manage my gamemodes, stay out.

That said, i'm still not too keen to take this if were going to take #416 to revert scoutknight too. There has been a lot of discussion to allow it toggleable, if that ever happens.

FortyTwoFortyTwo:
Well the fate has been decided from #416

Issue #429, opened 2025-05-10
Replace Scout's Stun on Rage with Slow and Single Jump


JuicyKaiser:
We had a conversation about this with multiple users, and even 42 agreed.

I suggest replacing the stun effect with a movement slow and restricting Scout to a single jump for a limited time. This would reduce his mobility without completely disabling him from doing damage and allow Hale to catch up to him. It would also make weapons like the Force-A-Nature and Atomizer more useful, as they can serve as escape tools under the new limitation.
The relevant code handling this can be found [here](https://github.com/redsunservers/VSH-Rewrite/blob/master/addons/sourcemod/scripting/vsh/base_boss.sp#L130-L140) and [here](https://github.com/redsunservers/VSH-Rewrite/blob/d22b177fd95a298ccc8ff088021c4113a80374df/addons/sourcemod/scripting/vsh/abilities/ability_rage_scare.sp#L5).

I've already asked a few people to implement it, but so far no one has been able to or had the time. Maybe someone else would be up for the task.

PR #432, opened 2025-05-11
Scout: Replace stun with a slow when raged

woisalreadytaken:
Addresses #429.

Just changing the stun flags isn't enough - slowdown stuns are really bad because you can get most of your speed back for free by moving diagonally. This takes a different approach: using a temporary attribute instead.

Movement speed is unchanged from before. Also disables double jumping and adds a few extra things because players wouldn't be getting stunned otherwise.

Also adds a `Slowdown` property to the scare rage.


addons/sourcemod/scripting/vsh/abilities/ability_rage_scare.sp
```diff
@@ -15,12 +17,14 @@ public void ScareRage_Create(SaxtonHaleBase boss)
 	boss.SetPropFloat("ScareRage", "Radius", -1.0);
 	boss.SetPropFloat("ScareRage", "Duration", 5.0);
 	boss.SetPropInt("ScareRage", "StunFlags", TF_STUNFLAGS_GHOSTSCARE);
+	boss.SetPropFloat("ScareRage", "Slowdown", 0.25);
```
FortyTwoFortyTwo, line 20: Does this default value affect all classes for a general rage? Even if it doesn't, is there really a purpose to have it here?

FortyTwoFortyTwo (approved):

PR #445, opened 2025-06-13
Remove Battalions Resistance from Hales rage

JuicyKaiser:
Hale is overperforming right now, and I feel like this is an appropriate nerf for him. It also makes the Kritzkrieg more usable now that it's not immediately shut down by rage.


woisalreadytaken:

JuicyKaiser:
Was not about to change it since I do not think Hale needs resistance at all on rage.

PR #446, opened 2025-06-14
Dragon's fury airblast cooldown

StarInAPond:
Dragon's Fury doesn't seem that much stronger to warrant 50% longer airblast cooldown, especially when it already uses pretty big number at base

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1016,21 +1016,7 @@
 				"block"				"pyro-primary-airblast"	//Phlogistinator don't have airblast
 			}
 		}
-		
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo (changes requested):
The entirely of Dragon's Fury bracket would need to be removed, both desp and `Tags_Airblast` as it'll default to the 12 seconds

StarInAPond:
I'm not entirely sure about Dragon's Fury having 12 second airblast as some people voices concerns, and DF really does have a longer range, so that value can be increased to 14–15 seconds if it will be problematic, or a simple removal of afterburn damage bonus, but 18 seconds is certainly an overkill.

FortyTwoFortyTwo (approved):

PR #448, opened 2025-06-17
Fast sandvich consumption speed

StarInAPond:
Bringing back old change from 2021 that's both beneficial and funny, makes consumables other than Dalokohs have a use


FortyTwoFortyTwo:
Bit of an unnecessary change to have

PR #453, opened 2025-06-17
Buffalo Steak real damage resistance

StarInAPond:
Since Buffalo Steak applies 20% damage vulnerability by default, VSH buff only results in -16% damage taken, which doesn't even help with damage threshold, you still die in 2 hits! This change applies real 30% damage resistance.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1571,7 +1571,7 @@
 				
 				"params"
 				{
-					"multiply"		"0.56"	//Multiply damage by 0.56 if under steak effect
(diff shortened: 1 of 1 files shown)


FortyTwoFortyTwo:
You can just set attrib 798 to 1.0 to remove the dmg vul, 0.56 isn't even the correct value to get back up to 30%

StarInAPond:
Are you sure? Steak uses one attribute to set all the variables, idk if a separate attribute will work, although I don't know how attributes really work internally

FortyTwoFortyTwo:
Steak currently have 798 as 1.2, that can be set to 1.0 to override its value, with 1.0 the same as default

FortyTwoFortyTwo (approved):

PR #455, opened 2025-06-17
Longer bonk

StarInAPond:
Defensive buff duration increase from 8 to 16 seconds


StarInAPond:
Reduced duration to 12 seconds

FortyTwoFortyTwo (changes requested):
Desc needs to be updated to match with 12 seconds

StarInAPond:
updated

FortyTwoFortyTwo (approved):

PR #456, opened 2025-06-17
Change Chargin' Targe resistance

StarInAPond:
Changes Chargin' Targe from active damage resistance you have to earn (requires minimum distance traveled + not hitting any teammates on impact) to passive 15%, to allow some interesting uses of Pain Train, Persian Persuader and Claidheamh Mòr. Will also allow Demoman to survive 1 hit without Booties or Eyelander. If new damage res proves to be problematic with certain bosses (50% fire + 15% general resistance against Pyro Car), fire damage resistance can be reduced.

As it is now, Chargin' Targe is outclassed by 2 other shields, I want Chargin' Targe to be the shield that enables some other less played weapons, as well as using it with popular choices, I suppose, there can never be enough damage resistance.

### DIFF
--- a/addons/sourcemod/configs/vsh/vsh.cfg
+++ b/addons/sourcemod/configs/vsh/vsh.cfg
@@ -1253,10 +1253,21 @@
 				"Tags_AddCond"
 				{
 					"cond"			"26"	//...give a defense buff (with crit resistance) to the attacker...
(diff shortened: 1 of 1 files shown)


StarInAPond:
Changed from passive buff idea to just increase the duration to 6 seconds

FortyTwoFortyTwo (approved):

PR #458, opened 2025-06-18
Degreaser change

StarInAPond:
As other Flamethrowers are focused on getting damage from their primary fire or afterburn, the purpose of Degreaser is to be a combo weapon. Faster holster speed will help it to combo with secondaries or melees, and 2 second lower airblast cooldown will serve a support role. Removed slightly lower afterburn to use normal tf2 value to empathize combo playstyle.


woisalreadytaken:

PR #461, opened 2025-06-18
Claidheamh Mòr

StarInAPond:
Increases Claidheamh Mòr charge time bonus by 1 second.


StarInAPond:
I was wrong, it's actually only +0.5 seconds compared to default, which is good enough

PR #462, opened 2025-06-19
Warrior's Spirit nerf

StarInAPond:
Small adjustment before previous buff even went live, instead of healing 45 hp more, Warrior's Spirit should have softer damage vulnerability.


FortyTwoFortyTwo (commented):
I don't quite see myself the needing of needing both positive changes, just lower the 120 HP on kill if its problematic

PR #475, opened 2025-10-26
Remove Outline

eno26:
Jesse Pinkman, you're a hard man to find, anyway, from the community feedback that I've gathered, it appears that no one ever was a big fan of this "thing", most of the time engineers get targeted by it (because they have the easiest method of dealing damage). I'm not really sure what the outline was meant to achieve, a lot of great combat situations get ruined because of it, esecially 1v1s


woisalreadytaken:
Wasn't opposed to doing this but just commenting out code isn't the play. I added the `vsh_top_score_outline` convar to control this instead, and disabled it by default since this feature seems to be unpopular. I see the vision and it could maybe be improved later, it's just that seeing players through walls might be too much.

PR #487, opened 2026-08-09
Unrevert the removal of changes from qol_extras now that they were fixed for the proper sm version

KirillianAmu:
tl;dr = forgot the server's sourcemod version is old
had no access to vrlab to test on a live environment. so locally i used new version of sourcemod.

now that i do have access to vrlab, i grabbed the correct version and reproduced the error properly.
this now has been tested on vrlab and a local srcds server and functions as expected.


KirillianAmu:
im gonna resubmit this shit because i have no clue how github works and it somehow is trying to merge weird shit incorrectly.

